PR #66, opened 2020-06-08
adds secondary Dockerfile for parallel monolithic image

Mearman:
will need the following build rules added to docker hub:

image


Mearman:
turns it doesn't like the "/" in the tag names, should be replaced with "-"

NotExpectedYet:
Sweet dude!

PR #259, opened 2021-02-05
WIP - Printer map, printer groups collection and printer group API / UI changes

InfoMTB3D:
Please see discussion
https://github.com/NotExpectedYet/OctoFarm/discussions/258


app.js
```diff
@@ -204,12 +204,13 @@ mongoose
     logger.error(err);
   });
 
+const port = 1963;
 const databaseIssue = async () => {
   app.listen(4000, "0.0.0.0", ()=> {
```
InfoMTB3D, line 209: Hmm merge whoopsie: db is set to 1963 locally as I run multiple non-docker versions on the same PC.

views/layout.ejs
```diff
@@ -50,15 +50,14 @@
             <a title="List Layout Monitoring View" class="nav-link text-light text-center btn-sm <%= noPrinters %>" id="menuListView" href="/mon/list"><i class="fas fa-list"></i> List</a>
             </li>
             <li class="nav-item">
-
-            <a title="Camera Layout Monitoring View" class="nav-link text-light text-center btn-sm <%= noPrinters %>" id="menuCameraView" href="/mon/camera"><i class="fas fa-video"></i>
-                Camera</a>
```
InfoMTB3D, line 55: Ah my bad, our manager would not even consider using camera. That's why I deleted this. Should revert that change if we merge at any point.

NotExpectedYet: Aye I noticed whilst I was looking at the changes. You've done well to figure out the views anyway, those are a mess aha. 

Yeah I won't be able to merge the hard coded port, is there any reason why you don't just update it in the UI? I gather the ports initially used but just thought I'd check. I'm moving over to a .env file for the 1.2 update so won't be a problem in the future. 

I'll have a play in the morning. Will the groupings and stuff work with just an IP:PORT combo already? I've got 100 instances I can throw at it to test aha. 

Once I've got my head round it all without asking you 1000 questions then I'll comment my thoughts.

davidzwa: Dont worry about the port, I will reset that change soon as well. Wanted you to have a free look at the concept of it

InfoMTB3D:
image

Latest version

views/layout.ejs
davidzwa: @NotExpectedYet see the update below, still havent cleaned up the PR yet. I'm also hesitant to merge it as we have to think about how the user configures this.

The farm I connected it too is now hard-coded in the constants, so it is far from mergeable.

NotExpectedYet: Aye certainly, I need to get chance to check it out rather full on at the moment. Got time to get it mergeable. 

Will those harded coded stuff work for me for testing? 

I will have a road map soon though :)

davidzwa: Oh it will work fine, it's just that your printers wont show up unless they have the right port/group name (Port 80 to 83 within a 'block' of 4 and RowX_Y where X must be between 0-1 and Y between 0-7) to fit in the coordinates.

NotExpectedYet: How will I test this if I won't see any printers?

davidzwa: If you can, provide a group name to Row0_0 and make sure a printer has port x0, x1, x2 or x3 (80 and 81 for example, but also 71).

I'll make sure to update this later or configure it through some way. Just gotta think about that a bit :)

NotExpectedYet: Aye I'm wondering how grouping would work. 

This could be my mis-understanding but wouldn't updating the grouping feature to allowing for an Array be a better place to put that? It just seems weird to use the ip/port combination when I know a lot of users wouldn't even think of grouping their systems like that. Especially with loads of printers. 

I feel OctoFarm's current state is a limiting factor for a decent grouping system. As I should be able to create groups on Printer Format, Custom Groups, IP ranges anything really.

NotExpectedYet: I gather you would have done that if you could as you've already hooked into the Drag and Drop one nice.

davidzwa: Ah yeah the job I did it for just required me to go live asap due to budget constraints. Now it's running (working almost error-less so far) and now I can slowly adapt the implementation to do it properly, in my own time.

About the port usage: it came from the fact that I mass flashed a lot of Raspberry Pi 4 SD Cards, so the ports were always the same. We run 4 instances of OctoPrint on each Pi. That might explain "why ports?". The groups were... because they were available, I couldve gone for hostnames as well, but groups made more sense to me as they abstracted away the Raspberry Pi.

Just wrote a big chunk of text on our scenario and the reasons for this port/group localization, but deleted again as it didnt go anywhere. I'm just not sure how to go about this. xD 
The take-away is:
**Mapping multiple USB/ACM ports to printers on a Raspberry Pi 4 with 4 OctoPrint docker containers can be inconsistent as the USB ports can swap around order/index, or dont seem to be absolutely fixated on one printer. So we lose our reference frame.** This is why I'm also not sure ports are the way to go. I'd rather have HID's or something unique from the Prusa's (or whatever printers).

Just an idea, what about having a docker container which pro-actively pushes (meta-)data about serial devices to OctoFarm? This way we can improve OctoFarm's knowledge before it even reaches OctoPrint. That would be immensely powerful, would it not?

davidzwa: So a lightweight discovery service, which is configured to securely update OctoFarm on hardware changes. OctoFarm could also query it, optionally.

NotExpectedYet: Aye following more now. I run a similar setup although I use a VM to host OctoFarm and it's OctoPrint instances. I had pondered on taking that further with USB over IP as well because currently USB length limits that, unless you splice the usb ends on an ethernet cable but whose got time for that XD. 

Aye I get that totally, so currently my main user base is of the Single Pi Per printer ilk, I have 1 user I'm aware of using some hacked together USB over IP protocol and skinny clients, and there are a few who do your method and a few who do my VM method as well. 

Whatever we decide should play nicely with every setup. Your use case it works fine for as you pre-flashed the devices in batch, but I get a general mix of users with OctoFarm, some have Pi's inplace already and do not want to change that, and others are coming it from it fresh and get to plan out the farm a little more with that in mind. 

I'm not 100% sure myself, but what does seem to be sticking in my mind a lot is just expanding the groups feature to not be a text field and to an array although I haven't given it much thought. All those groupings could be established that way, default groupings could also be setup to save time on hooks we have available and we support every configuration possible for OctoPrint. 

Yeah the ports can be an issue, especially with marlin. You get some board manufacturers that don't give out unique serial codes on the USB id's as well... then they've all got the same. To be blunt I see that level of management outside of OctoFarm totally and more reliant on OctoPrint but heck your playing into my fantasy thoughts here with your comments on this. I envisioned a similar type of system to PXE but centered around OctoFarm and the Pi device. A friend of mine has already written a decent set of ansible scripts that could accomplish that and set the ports up. The only other attempt at something similar I've seen is a kube setup, it didn't get very far though due to the port limitations. 
I gather you've tried all the usual UDEV hacks? The only other way to update that I think would be compiling your own firmware and heck... that would be a lot easier with Klipper because you could script the majority of that far more accessibly than I think marlin may do, that maybe my lack of time with the later marlins like so don't quote me on that. You want some form of USB unique identifier for your method and annoyingly easiest way to get that is use different printers but that's no good in most farm setups. 

Hmmm I have thought about something similar, and yes it would be quite powerful but I think that could only be truly utilized if we did something to skip OctoPrints requirement. If you had other thoughts on that I'm happy to hear them. 

I think instead of replying to your email a video chat might suffice more ahaha. My keyboard is ace but I can't adequately type my thoughts and dreams of where to take OctoFarm. Honestly it centers very much around the idea of replicating a AD methodology / maybe PXE booting the printer hosts (do that on a bloody MCU and that would be ace). Maybe we need to write out own MCU? XD just dreaming and thinking out loud to see if i'm on the similar wavelength here.

davidzwa: I like the enthousiasm and I was indeed hoping to communicate by video call eventually. Just didn't want to claim too much of your time on the other hand. Up to you entirely, we can also first work on making your Patreon subscribers happy and **then** build awesome features 👍🏼 

So we have 2 topics:
- really grouping instead of labeling and filtering on-the-fly (I'll leave that for later, it's not too hard)
- network and device discovery (+connect/disconnect events)

Yeah I used some UDEV rules, and I decided it was still too manual, as I had to explain everything to non-tech-savvy peeps and that was UDEV rules was not gonna cut it especially if it involved SSH. 
So, you mentioned different DIY printer firmware types I believe (google is my friend here), do those firmware types have a similar or well-known command which responds with over a identifier or metadata about the printer - **once connected**? We could connect shortly, query that, cleanup (to become static data) and hash that result and thus create a UUID ourselves and finally close the socket... just thinking out loud. This could mean some printers are, and some are not compatible with the aforementioned discovery service (when they dont have anything to identify them with like S/N or hardware description). We can't cover everything!
Oh btw, writing code for an ESP8266/Teensie/ATTiny to Intercept serial comms and proxy the comms with a unique one-time-only UUID... should be possible... I think? I imagine a small usb cable passthrough connector (exactly like a FTDI programmer but then with resettable VID/PID). It's just that we've got another device to setup, more costs per printer... it's backup option definitely for nasty firmware which does not have proper unique HID's.

Personally, I don't think it's smart to cover the area where OctoPrint shines (OS wise or Network Exec), I'd rather write a plugin for that or append a separate service on the same machine to push update OctoFarm and then have OctoFarm setup OctoPrint (a true puppet master). I have too much respect for OctoPrint to be able to dare dream replacing it or bootloading it remotely.

NotExpectedYet: I'd prefer it, I'm too easy to distract and forget to turn notifications off. If I know a time is coming up, it's easy to organise around that. If that is easier for you then that is totally an option. 

UDEV is a pia and I agree totally not for non-tech-savvy people, it's annoying for me and I'm tech-savvy. 

Personally, I'd do a udev rule specified to the port for your situation. I think it's the cleanest solution but there isn't going to be anything fired back to OctoFarm. We just give our printers crazy names from our favourite shows here. 

So the one I was specifically referring to was Klipper. Klipper is a special case compared to marlin, You have a smart host controller (klippy system), then you have the klipper firmware on the MCU. When you flash and install the firmware on Klipper (all cli) you can designate the specific USB id you'd like to use. You can also change the profile and serial? I forget, but the 0x78 / 0x99 codes that they pop up with in lsusb. So you wouldn't have to write anything extra around that. You could just detect the usb device been plugged in tell OctoFarm just for example, and OctoFarm could send the config that get's flashed. 

I use it myself personally as flashing marlin is a pain, and also configuring it afterwards is problematic. I just update a .config on my VM for this and then restart the controller. It's more programmer friendly, where as I gather marlin is more 3D printer person friendly. 

No couldn't agree more with that one. That's the beauty of OctoPrint you can just write a plugin for it. I have had thoughts but the only thing I'd remove in a similar scenario is the web interface and by that point, what's the point because Gina does good work and it's still in my opinion a handy fallback for OctoFarm to just get out of the way. Ansible would be totally the way to go with that as well, you got the whole hog of managment possibilies too.

davidzwa: Not ignoring the topics around Ansible and managing remote/device deployments, but I'd like to propose thinking about the grouping again as it's 'lower hanging fruit'. I think we'll eventually get to the point of managing IO and their firmware again, as it's all PIA without proper management.

What do you think of grouping printers by linking to a separate db collection (in this PR) and would we like distinct sets or sets with overlap?
Distinct: many printers have one (*..1) groups 
Overlap: many printers have many (*..*) groups  ==> can be powerful but chaotic to make 100% clear if done incorrectly in the UI

Yeah I'm a rel-db guy xD I'm excited to finally learn NoSQL db's like MongoDB

NotExpectedYet: I feel they should be split from this anyway, I was just jamming aha. The groups to me seems the simplest way to skin this task is ignore the automation for now. 

I like adding printers into their own collection yes that's fine, They can be linked via the ID's generated from the printers database. 

I honestly think we would be limiting ourselves sticking to the distinct option, especially if we want to shove in the automation for the things you mentioned above. Maybe put some buttons in that allow automatic grouping on command, or set up some default groups. You could easily group via the profile to find printers with similar XYZ sizes. Just ideas like. 

AHAHA oh NoSQL will be a breeze walk for you compared to MySQL and such. Dump and object, pull and object. It's great.

davidzwa: Is it smart to hop onto the private branch with an org? Otherwise db migrations might be a pain, but that's my XP from MSSQL/Postgres talking.

I'm willing to hop onto the changes to UI, and adapt the map view, if necessary.

NotExpectedYet: > Is it smart to hop onto the private branch with an org? Otherwise db migrations might be a pain, but that's my XP from MSSQL/Postgres talking.
> 
> I'm willing to hop onto the changes to UI, and adapt the map view, if necessary.

I may be mis-understanding that as I don't get how the database would have any effect on where the git is located?

davidzwa: Ah I was expecting you to have a lot of migrations and changes in the new version. I can also just do it here no prob!

NotExpectedYet: > Ah I was expecting you to have a lot of migrations and changes in the new version. I can also just do it here no prob!

The new version will migrate the collections over because there's stuff I wanna improve in the mongodb side of things where my inexperience lead to not keeping the schemas correct, and adding new settings is currently a pia. 

That will be all handled on the server though, and if your commit makes it in before I get the backend re-written then I can just add that into the migration path. It's better that groups be in their own collection anyway, rather than me dumping it all into the printers section like I did previously. 

My current plan for that is as follows: 

- Get the website updated to vue (currently on with, whilst prepping 1.1.11 bug fix)
- Migrate the current client over to vue (release on private github for patreons to try out)
- Make the server backend upgrades and modularisation 
- Update the client to follow the new API endpoints and what not
- Bring in something to document the API endpoints

NotExpectedYet: I'll jump you in before the "Make server backends" and get a project setup with what needs tackling if that's alright with you? 
Oh and btw, when this is all sorted your welcome to an OctoFarm T-Shirt as a thank you for all the work. Plus... If you'd be game, I will certainly need help (as in business partner) when it comes to setting up first the OctoFarm-Relay on OctoFarm.net, will relay your web interface so you can access remotely from your local install. Then also... OctoFarm-Online, I'm gonna sit down tomorrow morning and reply to your email properly anyway, and I'll detail these plans for you there. Just an offer anyway, I will more than likely be setting that up in my spare time unless my patreons keep exploding like they have done past few days.

davidzwa: Haha I prefer to keep a PR technical ;) You know how to reach me about those details.

I'll work on the group schema now

davidzwa:
Status of PR:

- [x] Added map to UI as PoC (still with constants and group-name + port to construct coordinates)
- [x] Add PrinterGroup collection to backed
- [x] Add mocha domain level tests for PrinterGroup validation

views/layout.ejs
davidzwa: https://www.npmjs.com/package/swagger-ui-express

For auto-API documentation. I use this a lot for .NET Core based API's, so it's really well-known. This library implementation I'm new to.

NotExpectedYet:
I think our linters are fighting

davidzwa:
Hmm could be, I work in Webstorm which automatically takes your ESLint file. I think somewhere tabs get converted to spaces by pressing Ctrl+Shift+L. What IntelliJ software are you using?

NotExpectedYet:
> Hmm could be, I work in Webstorm which automatically takes your ESLint file. I think somewhere tabs get converted to spaces by pressing Ctrl+Shift+L. What IntelliJ software are you using?
> 
> ... Ah the same I see. We should definitely align, dont want to spam github with whitespace changes xD

Aye, I'm webstorm as well. I've just updated my linter to use prettify again in the dependencies and it "should" carry over on your install.

NotExpectedYet:
I've never shared linting stuff so just let me know if it aint the case or if I need to make any changes

davidzwa:
Yeah linting config should definitely be shared. There's also JS tools which can be run on Github to check it for you, so a config is nice.

NotExpectedYet:
> Yeah linting config should definitely be shared. There's also JS tools which can be run on Github to check it for you, so a config is nice.

Another avenue I need to spend more time on ahaha 

Well mines updated on the latest dev branch so just let me know if there's still issues or tweaks to make to it. Honestly as long as it formats quote issues and lines I'm never that bothered personally what rules we specifically use.

davidzwa:
https://github.com/marketplace/actions/lint-action

Oh btw, I'd like to run Mocha/Karma tests on an action as well. You ok with that? This is a way for me to understand an app better + it can really prevent core feature bugs.

NotExpectedYet:
> https://github.com/marketplace/actions/lint-action
> (It's one click away ;) )
> 
> Oh btw, I'd like to run Mocha/Karma tests on an action as well. You ok with that? This is a way for me to understand an app better + it can really prevent core feature bugs. My style is to add 'core tests' only, so not like 1000 useless tests. Just a couple of necessary ones.
> https://docs.github.com/en/actions/guides/building-and-testing-nodejs#starting-with-the-nodejs-workflow-template
> Also 0 effort == yay!

Go for it :) There's some weird UI oddities people are getting and I can only imagine its some mis-match somewhere.

NotExpectedYet:
Do a seperate PR though on dev, I'm just building out a printer statistics modal and I'm releasing 1.1.11 so depending how quick you are might get it up to master. Saying that I have no idea how long that stuff takes. It will be helpful to read it though for me because I can't currently get my head around unit testing.

PR #281, opened 2021-02-13
Add old babel eslint package 

davidzwa:
Temp fix

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "octofarm",
-  "version": "1.1.9",
+  "version": "1.1.10",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


NotExpectedYet:
Sorry man, pulled that before checking properly again.

Pull latest dev, had bloody eslint in ignore... wondered why you didn't see that.

module.exports = {
  root: true,
  env: {
    browser: true,
    node: true,
  },
  extends: ["plugin:prettier/recommended"],
  plugins: ["prettier"],
  parserOptions: {
    ecmaVersion: 2020,
    parser: "babel-eslint",
    sourceType: "module",
  },
  rules: {
    "prettier/prettier": "error",
    "max-len": [
      "error",
      {
        code: 300,
        ignoreComments: true,
        ignoreUrls: true,
      },
    ],
    "global-require": "off",
    quotes: ["error", "double"],
    "no-console": process.env.NODE_ENV === "production" ? "warn" : "off",
    "no-debugger": process.env.NODE_ENV === "production" ? "warn" : "off",
  },
};

Issue #284, opened 2021-02-13
Typo in System panel (.snapshot instead of .snapshots)


davidzwa:
### To better help the developer understand your issue, please fill in the information below. Failing to do so will waste the developer and your own time in sorting the issue. Thanks!

**Describe the bug**
Typo in property accessor `res.history.snapshot.onComplete`

**To Reproduce**
Look at:
```
          document.getElementById("snapOnComplete").checked =
            res.history.snapshot.onComplete;
```
Place debug breakpoint
Reload
See that `snapshots` looks like, but is not equal to, `snapshot` 

**Expected behavior**
No errah

**Screenshots**
![image](https://user-images.githubusercontent.com/6005355/107862873-4e001780-6e50-11eb-82bd-dffe21e37d7e.png)

**Versions (please complete the following information):**
Latest

**Additional context**
This is what typescript can prevent jk :)

NotExpectedYet:
Good catch! I think a priority might be getting that webpack sorted what do you think?

Issue #283, opened 2021-02-13
UsageOverTime does not have index 0 when history is still empty


davidzwa:
Latest DEV version (partially with webpack, but it's not related) ;)

![image](https://user-images.githubusercontent.com/6005355/107862596-fbbdf700-6e4d-11eb-926a-4ebd47418b71.png)

dashboardRunner.js
```
  let usageOverTime = historyStatistics.history.usageOverTime;
  
  ...
  {
        seriesName: usageOverTime[0].name,
  }        
```

Couple of other places as well, please check thorougly!

davidzwa:
Will be fixed in webpack release. Small ternary operator fix.

NotExpectedYet:
Hmmm yeah that would probably break thinking about it. 

Another good catch. I've got a few hours this morning so I'll take a look at the webpack PR.

PR #286, opened 2021-02-13
Switches to webpack, fixes minor frontend bugs

davidzwa:
- [x] Builds runners and most vendor assets with `webpack --watch`
- [x] Builds source maps
- [x] ~Fix GridStack looking weird #288~
- [ ] Builds CSS
- [ ] Can run lightweight dev watch within `app.js` when in development mode
- [ ] Has custom production version of both JS/CSS/images with optimization turned on to full

Also fixes:
- [x] #283 dashboard index check
- [x] #284 system `snapshot` typo
- [x] Prefers MONGO environment var (dotenv) over `config/db.js` URI

Can I skip processing images? I don't really see the point.


NotExpectedYet:
Yeah shift the images, it's all local anyway and they're not originally massive. There's an issue with the current gify-img thing on gulp anyway so it's not been doing it for a while.

davidzwa:
Abandoning due to bootbox being a ******* mess.

My webpack skills are not pro enuf for that

PR #290, opened 2021-02-14
Printer statistics

NotExpectedYet:
Current work on the printer_statistics branch to merge over. Rest of information is required from a new API endpoint /api/system/info OctoPrint.  Tested on local branch and seems fine at the moment, information generation uses same process as historyClean so should be robust enough. 

Note: requires history to generate. 

ToDo: 
 - [x] Check with 0 history. 
 - [x] Add note on screen stating generates when history has been. 
 - [x] Create check for API endpoint 
 - [x] Generate API information through printer clean, pass through the object...
 - [x] Display select stats on the UI
 - [x] Add a firmware column to Printer Manager


NotExpectedYet:
Gather we are best re-compiling the client after merge. Didn't think of that.

davidzwa:
What I did was click ignore on those files and untrack those files. That way they are kept whilst being ignored.

Soft reset the commit with these files, ignore + untrack in the changes in gitkraken and you should be good. 
Finally it requires a force push to appear here ofc.

NotExpectedYet:
Ah much easier thank you! 

Will sort it.

NotExpectedYet:
Sorry made a goof somehow with that reset, wouldn't work. Learned more at least, won't commit those again. Merged it so I can sort development and get testers on it.

NotExpectedYet:
I've sorted it but pushed a few things straight to dev because I've got to go to work soon and I somehow deleted your previously merged changes to fix those two bugs / gitignore. Should know what I'm doing now.

NotExpectedYet:
I wonder if we should use a staging before development? Currently dev / master get generated by docker on the update of the branch. Was thinking if we collate everything in staging before, we can do the compile there and push all to development. Then onto master. What ya think?

davidzwa:
@NotExpectedYet I think we should try and stick with development ;) It's what its there for.
Merging PR's should be a point of reflection: don't hit the merge button before all the changes files have at least been skimmed through. It might really surprise you how much you might catch still.

Oh and build checks: we should have the NodeJS build/test action enabled as a check, I think it's not enabled automatically because development is not the default branch.

NotExpectedYet:
> @NotExpectedYet I think we should try and stick with development ;) It's what its there for.
> Merging PR's should be a point of reflection: don't hit the merge button before all the changes files have at least been skimmed through. It might really surprise you how much you might catch still.
> 
> Oh and build checks: we should have the NodeJS build/test action enabled as a check, I think it's not enabled automatically because development is not the default branch.

makes sense, I'll get rid of the normal one. My confusion was like I've just had to do. The client needs re-compiling to get the latest changes, so I just switched over and pushed the compiled code. Totally agree with that, I spotted a thing or two on this one. 

I'll have a look at settings for that now.

NotExpectedYet:
Done development is now default. Thanks for that.

PR #316, opened 2021-02-20
313 server api migration - user registration

davidzwa:
- [x] Adds user registration
- [x] Fixes gulp for new situation
- [x] Use flash
- [x] Add username existence check and min length username/password


davidzwa:
@NotExpectedYet let me know what you think, its not a lot so pretty easy to review

PR #332, opened 2021-02-26
WIP Client connection substate

davidzwa:
Start working on draft singleton for client connection substate.
Will be inflated later and roofed under a list of providers, which are initialized at startup and rehydrated with database/persisted last-known state if available.

- [x] Use yarn in github e2e workflow
- [x] Fix e2e tests (both in Webstorm, in NPM v13 docker and on gitty)
- [x] Fix mongo in-memory with TypeORM (big yay)
- [x] Add OP API schema generator tool (very cool)
- [x] Add OP connection state prototype (almost 100% done for 1 connection) + state tests


davidzwa (commented):
I checked it all. LGTM from my perspective, some WIP stuff which is not blocking and works fine.

NotExpectedYet (approved):

PR #340, opened 2021-02-28
Websocket client sub-state

davidzwa:
- [x] Authenticate with auth/login => sessionKey+username => websocket
- [x] Wrap up state in 200 OK mode
- [x] Connect succesfully receiving history/timelapse/connection messages
- [x] Transform deserialized message dto's from Websocket messages incoming
- [x] Connect and succesfully test receiving stream
- [x] Transform event message types to recognize and model payloads
- [x] Setup message dto's from Websocket messages outgoing (throttle/auth)
- [x] Test websocket client with OP auth flow
- [x] Implement reconnect and backoff
- [x] Make connection state react to events (websocket disconnect/retry, API spotty, power-down or auth wrong)


davidzwa:
image
Just me showing off

davidzwa:
image
Now with all message types I could think of (could be more)

davidzwa:
Halted by some design architectural thinking. Iterating and will continue soon

davidzwa:
Gonna replace backoff and regen later. This is the ugliest way to implement it, but cant do everything in 1 PR.

NotExpectedYet (approved):
This is looking epic. 

I just have one comment and it's regarding how we will incorporate multiple printer systems as I'd like v2.0 to ship with a few. 

I was thinking something along the lines of having an "OctoFarm State" and then your OctoPrint ones you've done here would feed back to that, our client would be pulling from the OctoFarm State. I gather this is currently possible with the code, it looks like it. I'm all ears about a cleaner solution to that too as it was just my initial thoughts on it some time ago. 

Needs discussion anyway about what we want to specifically show to the user and how to handle the difference's unless you've already thought about it. I'm going to keep scanning over this code anyway and get more familiar with it.

Issue #23, opened 2020-04-06
Camera doesn't apply multiple rotation settings


NotExpectedYet:
Camera currently only applies the top rotational setting and doesn't apply both.

NotExpectedYet:
Just on views from my tests

dburr:
In my case, "Camera currently only applies the top rotational setting" does not seem to apply. I have my RasPi camera mounted upside down because of cable management issues, and so I have to have both "flip horizontally" and "flip vertically" enabled. However OctoFarm shows the image correctly flipped vertically, but NOT flipped horizontally as it should be (so, in effect, the image appears mirrored.) So in my case it's the SECOND setting (flip vertically) that is being correctly applied, NOT the first (flip horizontally.)
![Screen Shot 2020-04-12 at 2 13 47 PM](https://user-images.githubusercontent.com/1237762/79080085-ec1ed080-7cc7-11ea-85b0-65a185c6f53b.png)
![Screen Shot 2020-04-12 at 2 13 52 PM](https://user-images.githubusercontent.com/1237762/79080087-ed4ffd80-7cc7-11ea-8fcb-a1318b4ff6d7.png)

NotExpectedYet:
The title doesn't say that, I'd edited it... As it says it applies the first, not multiple. Will be fixed in a later release.

dburr:
> The title doesn't say that, I'd edited it... As it says it applies the first, not multiple. Will be fixed in a later release. Just because it's first there doesn't mean it's first in the code as well. I know where the bug is but thanks for the additional info.

OK, thanks for the reply. Looking forward to it -- viewing my (printer's) world horizontally flipped is a bit disorientating ;-)

NotExpectedYet:
No worries :) Yeah can imagine, I got lazy when I was doing the views if I'm honest. I remember thinking, no one will need both aha. Certainly be sorted for the release of 1.1.5 though. I don't have any work arounds at the minute either but I'll see if I can push that to master within the week for you as  a few other people have mentioned it.

NotExpectedYet:
@dburr 1.1.5 is on development currently and fixes that. You can check the changelog to see if you want to change over. Currently there's some stat's missing but everything else is in there.

jcassar:
[...]

PR #347, opened 2021-03-04
Development 1.1.12 hotfix

NotExpectedYet:
See Project #23 

All completed.


server_src/lib/serverCommands.js
```diff
@@ -8,13 +8,22 @@ var wol = require("wake_on_lan");
...
+      //Attempts to find the named service
+      const { stdout, stderr } = await exec("pm2 pid OctoFarm");
+      if (isNaN(parseInt(stdout))) {
+        throw "No process number returned";
+      }
       logger.info("stdout:", stdout);
-      logger.info("stderr:", stderr);
+      checkForNamedService = true;
+      // If the process had a integer pid succeeded then we are ok to restart, errors are caught by the try, catch
+      exec("pm2 restart OctoFarm");
```
davidzwa, line 22: Maybe introduce a small timeout as to prevent possibilities of rapid restarts. It's something I got a bit scared off using pm2... however stupid or irrealistic it may sound

server_src/routes/printers.js
```diff
@@ -131,7 +131,8 @@ router.post("/printerInfo", ensureAuthenticated, async (req, res) => {
     res.send(printers);
   } else {
     const index = _.findIndex(printers, function (o) {
-      return o._id === id;
+      //Make sure ID's are both strings to stop recursion issues
+      return JSON.stringify(o._id) === JSON.stringify(id);
```
davidzwa, line 135: Stringifying objects... ok ehm wow sure. It is indeed a fix. Object.Is(obj1, obj2) is as well.

NotExpectedYet: I could get rid of the === and put == but I feel that was a worse solution.

server_src/lib/serverCommands.js
NotExpectedYet: I did consider that tbh with a setTimout, also could do with diabling the button client side too after a press. Stops any possibility of one then.

server_src/routes/printers.js
NotExpectedYet: I feel that can be looked at at a later date, currently file manager isn't working which is a biggy. Although I do agree, but I haven't had chance to update my old habbits yet.

davidzwa (changes requested):
Gotta reject on the JSON.stringify comparison as it is very prone to fail and not fix bug you intended to patch, see the comment. The rest is all good or optional.

server_src/lib/serverCommands.js
davidzwa: Ok feel free to do or to skip.

server_src/routes/printers.js
davidzwa: Ok as a promise from my V2 side: I will fix it on the API and ensure everything is string, ok? Then Im ok with merging.

davidzwa (approved):

davidzwa:
Based on the promise of fixing the ObjectID serialization in V2, LGTM

server_src/routes/printers.js
NotExpectedYet: Well the way I see it currently is that call is from farmPrinters and that's in state. That's a mess XD, I'm relying here on the fact that that always get's dumped with an ID which in this situation it's going to be because if it hasn't something else will have failed before this always get's called. 

I wouldn't see if as an express issue personally with how the code is laid out but happy to be proved wrong. It needs pulling out of the database as a string permanently but I feel there's somewhere in the code where it's not getting done and finding that will be difficult. 

I to me have to options in this situation and that's make sure it's a string here as a "patch" only until we rip apart the state file and the rest, as you'll know when you update and object that's copied it also updates the original. Or let the double "==" handle it and call it a day?

server_src/lib/serverCommands.js
NotExpectedYet: I gather the setTimeout would be saved in a variable and checked for its existence here? Our lass is yogaing so can look at that now. I'd rather not people rapid fire, I'll also do the client changes too and make sure that is disabled until we have a response from the server.

davidzwa (approved):
LGTM eet smakelijk

PR #356, opened 2021-03-07
OctoFarm Installation / Manage Script

NotExpectedYet:
Make startup "service" wrapper with a JS Script to load, restart, stop, update OctoFarm. 

Allow control from the app. 

 - [ ] Initialise the application. - Install required dependencies, setup database
 - [ ] Allow control from within the application to Start/Stop/Restart and update the application
 - [ ] Possible spin up a endpoint so can be created through the web interface.


NotExpectedYet:
Could possibly include the below to get around any db changes someone might have made. 
`git update-index --assume-unchanged configbd.js`

NotExpectedYet:
Updated the update script for that. 

Let me know your thoughts.

package.json
```diff
@@ -7,6 +7,8 @@
     "prestart": "npm ci",
```
davidzwa, line 7: I used `npm ci --production` recently as this is now also installing things from dev-dependencies. Especially mongo-memory-server is a big payload of 280MB which is useless to download. Can you add the --production flag?
Also shall we move nodemon to non-dev dependencies? This way people can use it if they cannot use pm2. Did I mention the flashing issue with pm2 did not occur on MTB3D's PC?

package.json
```diff
@@ -7,6 +7,8 @@
...
     "start": "pm2 flush && pm2 start app.js --name OctoFarm -e ./logs/pm2.error.log -o ./logs/pm2.out.log --time",
     "stop": "pm2 stop OctoFarm",
+    "restart": "pm2 stop OctoFarm && pm2 start OctoFarm",
+    "update": "git update-index --assume-unchanged ./config/db.js && git pull && npm ci && npm run restart",
```
davidzwa, line 11: Since I updated instructions on OF readme to use `.env` it seems more logical to use .env for database now. 
We can write a startup script to create the .env file when it doesnt exist which copies the MongoUri from the config/db.js file, so people dont have to worry about that. In case .env already exists and has a MONGO variable, we can leave it safely as is.
I think this npm script kinda stuff is a bad idea, especially since it does not fix the problem that the file still causes git changes unless the update command is called once... please no. Instead we need to have something more robust and self-healing rather than put into custom git commands, which are quite easy to get wrong.
Use gitignore properly: ignore, untrack, and self-heal with a smart startup js/ts script which is called by our application.

davidzwa (changes requested):
We should think a bit more about the proposed changes as I can think of quite some scenario's which these will cause more confusion and inconsistencies than intended. Instead, I'd like to suggest a separate startup script or tool, which calls the right npm/git or pm2 commands. This looks like a CLI and I think that is what we should move to soon anyway... a tool which helps people run, update and restart OctoFarm safely or abort with the proper error message + tips, don't you agree?
A script gives us the breathing room to perform actions in the way we want it to, NPM scripts and `&&` are too easy to bail out with no proper error messages i.m.o.

package.json
NotExpectedYet: Yes noticed myself. Will do

I can only gather it's the mis-match with pm2 local/global. It's the only thing I can put it down to at minute that I haven't personally tested.

package.json
NotExpectedYet: That wasn't the intention. The purpose of this was to lay the ground work that will support users now and allow us breathing room to update and change how we see fit in the future.

PR #362, opened 2021-03-13
Compile and Build a .deb package for Ubuntu TESTING

NotExpectedYet:
pre-installation requirements:
1. `sudo apt install dpkg autoconf fakeroot jq`
2. I had to install node-deb globally, might not be required on all systems. 
    - `npm install -g node-deb`

Currently draft until completed the below: 
 - [ ] Migrate to use the pkg linux file that's generated
 - [ ] Add internal checks for pm2, systemd services
 - [ ] Migrate restart functions to run the specified service

Not as important: 
 - [ ] Figure out how to create a hosted repository on octofarm.net
 - [ ] Create Github Actions workflow based on master releases to build and add to release packages
 
 CAVEATS to resolve:
  - [x] incorrect  ~~Seems we can only specify 1 location to build from. Currently this just means can only build when no node_modules exist as they are built on installation~~
  - [x] ~~Fails on start due to not been able to create the correct logs folder. Relative PATH issue, weird because I see the files created in /usr/share/octofarm/bin/logs~~ This works with permissions updated. Needed to set on prestart
  - [x] ~~Local MongoDB only~~ config can be edited in /usr/share/OctoFarm/app or .env can be added in same place
 
 Build instructions:
`node-deb -- app.js views/ server_src/ images/ config/`


NotExpectedYet:
pm2 is proving problematic whereas the normal node command runs fine. 

Thinking about sacking pm2 off but means detecting the usage of the systemd script and restarting that way. 

Would just be a fallback to `sudo systemctl restart octofarm` if so but would require elevated privileges inside the app to execute.

package.json
```diff
@@ -113,5 +114,20 @@
       "client_src/*"
     ],
     "delay": "2500"
+  },
+    "node_deb": {
```
davidzwa, line 118: Shall we expose this into a file?

NotExpectedYet: It didn't seem to be an option with node-deb but I'll check as I'd much rather.

davidzwa: Ok nvm I think you're right. No worries!

NotExpectedYet:
Going to close this as well, not hard to reproduce once we have finalised stuff regarding packaging.

PR #375, opened 2021-03-17
Ignore background image and copy a default if not present 

davidzwa:
- Copy a default file to images/bg.jpg to keep it untracked on github


NotExpectedYet (approved):
Sick man love that!

PR #376, opened 2021-03-17
Octofarm self-update/restart functionality

NotExpectedYet:
OctoFarm internal update command. 

 - [x] Setup buttons on client for Update / Force Check
 - [x] Create route for OctoFarm client to call
 - [x] Create initial serverCommands function for updating OctoFarm. It needs to: 
      - [x] Check for the git folder as this will only work with git installs currently. 
      - [x] Run the git pull command 
      - [x] Check for "Already up-to-date status from git pull
      - [x] Check for local changes to git folder and fail on update
 - [x] Hook in a function for the force check for an update with an API endpoint so client may call at their whim. 
 - [x] Make the AmIAlive check deactivate or activate the client "Update" button when update available / not. 
 - [x] Add UI information from the update check call inc. node information and the rest from latest commit. 
 - [ ] Add in smoke test for the serviceInformation in index.js
 - [ ] Add in smoke test for /server/update/check

 Left as draft currently for initial feedback.


NotExpectedYet:
@davidzwa 

Just an initial draft here, put check marks on still to do and happy for input if you forsee any issues. I think when this release is pushed we will have solved all the "git pull" issues so it might just be worth adding a hard fail if any do crop up as they will be on the user.

NotExpectedYet:
And I forgot the changelog 🤣

server_src/lib/serverCommands.js
```diff
@@ -25,6 +30,45 @@ class SystemCommands {
...
+    try {
+      // Check to see if current dir contains a git folder... hard fail otherwise.
+      let doWeHaveAGitFolder = await isGitSync("./");
+      if (!doWeHaveAGitFolder) {
+        let notGitRepoError = "Not a git repository, manual update required...";
+        messageForUser = notGitRepoError;
+        statusTypeForUser = "error";
+        throw notGitRepoError;
+      }
+      logger.info("Root location contain a git folder: ", doWeHaveAGitFolder);
```
davidzwa, line 48: Log really necessary? Just a reminder that logging in general can be taxing on CPU / file IO.

server_src/lib/serverCommands.js
```diff
@@ -1,9 +1,14 @@
 const fs = require("fs");
+const path = require("path");
 const util = require("util");
 const exec = util.promisify(require("child_process").exec);
 const Logger = require("../lib/logger.js");
 const logger = new Logger("OctoFarm-Scripts");
```
davidzwa, line 6: logger name does not correspond file name, that intended?

NotExpectedYet: It's me segregating it, this file needs splitting up I think. Server commands != The scripts feature. The server commands need to be in the OctoFarm-Server.log file and Scripts could either got in the API / their own. 

I don't generally group by file name for logging I split by system area so when an issue arrises I can get a single log file specific to the area of the issue, this doesn't really follow that since including the system commands so thanks for pointing that out.

server_src/lib/serverCommands.js
NotExpectedYet: Not really as the rest of the function is more important I'll remove this in the next PR.

NotExpectedYet: I'll keep that in mind for the server.js split as well as there is an ungodly amount of stuff logged. I propose we just log errors or anything relavant to connections in files then and completely reduce the IO but I'll add that to the server.js projects bots as it's not relevant here.

client_src/js/serverAliveCheck.js
```diff
@@ -33,7 +33,13 @@ function checkUpdateAndNotify(updateResponse) {
     } catch (e) {
       parsedStorageReleaseInfo = null;
     }
-
+    // If the update button is available, we are on the system page. React to available updates accordingly.
+    let forceCheckForUpdatesBtn = document.getElementById(
+        "checkUpdatesForOctoFarmBtn"
```
davidzwa, line 38: Repeated string literal, should be made constant. Or you can save the queried element. Your choice.

server_src/routes/settings.js
```diff
@@ -127,6 +132,11 @@ router.get("/server/update", ensureAuthenticated, async (req, res) => {
   let octoFarmUpdated = await SystemCommands.updateOctoFarm();
   res.send(octoFarmUpdated);
 });
+router.get("/server/update/check", ensureAuthenticated, async (req, res) => {
```
davidzwa, line 135: Add API smoke-test. Look at supertest usage (test already present in test folder)

server_src/routes/index.js
```diff
@@ -490,6 +494,13 @@ router.get("/system", ensureAuthenticated, async (req, res) => {
     systemInformation,
     db,
     dashboardSettings: dashboardSettings,
+    serviceInformation: {
```
davidzwa, line 497: Add API smoke-test. Look at supertest usage (test already present in test folder)

views/system.ejs
```diff
@@ -26,6 +26,12 @@
                                             <%= systemInformation.warnings.message %>
                                         </div>
                                     <% } %>
+                                    <% if(serviceInformation.isPm2 !== 'undefined' && !serviceInformation.isPm2){ %>
+                                            <div class="alert alert-warning" role="alert">
```
davidzwa, line 30: Smoke-test with supertest for containing 'Your system is not running with PM2'.
https://www.codota.com/code/javascript/functions/supertest/Response/text

PM2 can easily be mocked by overriding (bit dirty yeah) the environment variable which is checked in `isPm2()`

views/system.ejs
```diff
@@ -73,28 +79,48 @@
...
-                                            <!--                                    <p><i class="fas fa-code-branch"></i> Latest Version:                                     <a title="OctoFarm Version, will take you to github changelog" class="text-light"-->
-                                            <!--                                                                                                                          href="https://github.com/NotExpectedYet/OctoFarm/releases"-->
-                                            <!--                                                                                                                          target="_blank"></a></p>-->
-                                            <p class=""><i class="far fa-clock"></i> Timezone: <br><span class="text-light" id="timeZone" style="font-size:0.8rem;">
+                                                    </p>
+                                                    <p><i class="fas fa-code-branch"></i> Version: <a title="OctoFarm Version, will take you to github releases page" class="text-light"
```
davidzwa, line 89: Not the latest release? We get an URL with the release response :)

views/system.ejs
NotExpectedYet: Check for the PID inside the isPm2 function maybe? would be much harder to override and can be then backfilled to the restart function.

views/system.ejs
```diff
@@ -73,28 +79,48 @@
...
+                                                        <p class="text-light">Powered by <i class="fab fa-node fa-2x"></i></p>
+                                                    <% } %>
+                                                    <% if(serviceInformation.update !== 'undefined' && serviceInformation.update.air_gapped !== 'undefined' && serviceInformation.update.air_gapped) { %>
+                                                        <p class="text-light"> <i class="fas fa-air-freshener"></i> Farm is air-gapped</p>
+                                                    <% } %>
+                                                    <% if(serviceInformation.isPm2 !== 'undefined' && serviceInformation.isPm2){ %>
+                                                        <p class="text-light"> <i class="far fa-eye"></i></i> Running under pm2</p>
+                                                    <% } %>
+                                                    <% if(serviceInformation.isNodemon !== 'undefined' && serviceInformation.isNodemon){ %>
+                                                        <p class="text-light"> <i class="far fa-eye"></i> Running under nodemon</p>
```
davidzwa, line 118: Test for response content using supertest?

views/system.ejs
```diff
@@ -73,28 +79,48 @@
...
+                                                    <p><i class="fas fa-layer-group"></i> Service Information:
+                                                    <% if(serviceInformation.isDockerContainer !== 'undefined' && serviceInformation.isDockerContainer){ %>
+                                                        <p class="text-light"> <i class="fab fa-docker"></i></p>
+                                                    <% } %>
+                                                    <% if(serviceInformation.isNode !== 'undefined' && serviceInformation.isNode){ %>
+                                                        <p class="text-light">Powered by <i class="fab fa-node fa-2x"></i></p>
+                                                    <% } %>
+                                                    <% if(serviceInformation.update !== 'undefined' && serviceInformation.update.air_gapped !== 'undefined' && serviceInformation.update.air_gapped) { %>
+                                                        <p class="text-light"> <i class="fas fa-air-freshener"></i> Farm is air-gapped</p>
```
davidzwa, line 112: Test for content? Can be mocked, needs some extra work in mocking the air-gapped boolean I guess

views/system.ejs
davidzwa: I dont see why actually. What would be your reasoning for making `isPm2()` checking for PID?
Btw, I edited my comment above with the way on how to mock the env variable

davidzwa (commented):
Once you have the tests and they are in the green I will smash the approve button to bits.

LGTM code-wise.

views/system.ejs
NotExpectedYet: Good point actually wanted to show connection available as an else as well.

views/system.ejs
NotExpectedYet: Aye will check your code a bit more and add that good idea.

views/system.ejs
NotExpectedYet: The only reasoning behind that is it can't be mocked. It has to be running as a pm2 service to get back a response from pm2 itself. It's a certain, this is pm2 to me.

NotExpectedYet:
@davidzwa I was thinking it might be beneficial to get this live for 1.1.13 what do you think? We've got an automated base to change things going forward then. It can still work with pm2/git for now and we can update at a later date when/if the packaged versions are ready. 

Just thinking it might not be worth holding 1.1.13 back for packaged versions what you think?

NotExpectedYet:
I'd only need to finish the smoke tests plus I forgot the new system info stuff was sat in here which is desired

server_src/routes/settings.js
```diff
@@ -123,6 +128,15 @@ router.get("/server/restart", ensureAuthenticated, async (req, res) => {
   let serviceRestarted = await SystemCommands.rebootOctoFarm();
   res.send(serviceRestarted);
 });
+router.get("/server/update", ensureAuthenticated, async (req, res) => {
+  let octoFarmUpdated = await SystemCommands.updateOctoFarm();
```
davidzwa, line 132: try catch here, if you must.

server_src/lib/serverCommands.js
```diff
@@ -16,28 +22,75 @@ class SystemCommands {
...
     }
     return checkForNamedService;
   }
+  // This will need changing when .deb / installation script becomes a thing. It's built to deal with the current implementation.
+  static async updateOctoFarm() {
+    loggerServer.info("Update OctoFarm server requests");
+    let haveWeSuccessfullyUpdatedOctoFarm = false;
+    let statusTypeForUser = null;
+    let messageForUser = null;
+    try {
```
davidzwa, line 39: I advise not to try catch here, but let the caller do that (route?)

server_src/lib/serverCommands.js
```diff
@@ -16,28 +22,75 @@ class SystemCommands {
...
+    }
+    return {
+      haveWeSuccessfullyUpdatedOctoFarm,
+      statusTypeForUser,
+      messageForUser,
+    };
+  }
 }
 
 class Script {
```
davidzwa, line 83: - multiple classes in 1 file
- what do scripts have to do with system file-IO and update commands?

server_src/routes/settings.js
davidzwa: Oh btw, why does this have a GET method? Shouldnt this be POST?

davidzwa:
Package version still has problems of its own. Not gonna be ready soon, so PKG must be postponed. The update functionality is top priority. I'd prefer to ship it when it is robust though. I will test this functionality when home.
On my try-catch and exception comments above: they do not hold back approval, just would like to you to see them as prep-work for V2.

Can you add the API tests? They are super easy and simple to copy paste within server-check.test.js.
Here's what I would try to add:
```
describe("ServerUpdate Endpoint", () => {
  it("should not update when there is nothing to update", async () => {
     // Set update by overriding process.env.npm_package_version = ... 
     // call supertest with POST /server/update (do note that it should technically be POST, not GET)
     // expect something within response        
  });
  it("should call mocked update when there is an update", async () => {
     // Set update by overriding process.env.npm_package_version = ... 
     // call supertest with POST /server/update (do note that it should technically be POST, not GET)
     // expect something within response indicating the update action
  });  
    it("should not call update when this is a non-released version", async () => {
     // Set update by overriding process.env.npm_package_version = ... 
     // call supertest with POST /server/update (do note that it should technically be POST, not GET)
     // expect something within response indicating the update action
  });  
});
```

Additionally, we'd need something which mocks `SystemCommands.updateOctoFarm`.
This needs to be avoided in tests: `const { stdout, stderr } = await exec("git pull");`
Jest has a nice way to mock files by adding a __mocks__ folder
image
This is on the HistoryClean PR, so cant link code.
The test file then imports the mocked service like this:
```
jest.mock("../../services/history.service");
const mockHistoryService = require("../../services/history.service");
```
So as you might be able to see there is no `__mocks__` in the `jest.mock(...)` input. Jest assumes mocks to be placed in any subfolder with that name.

davidzwa:
@NotExpectedYet closing, as #420 supersedes by rebase and squash!

PR #390, opened 2021-03-24
Bugfix/filament manager cleanup

### LINKED ISSUE #346: Filament Manager sync just one way?
**Describe the bug**
Not sure if it's the actual design, but by synchronizing the filament manager with one octoprint the content from Octoprint is synced into the octofarm manager. If a new spool is added within Octofarm, those information are not synced back to the Filament Manager of the connected Octoprint. Also changes of the used filament are not synced to other octoprint instances filament manager.

Would be nice to have the content always synced between Octofarm and ALL connected Octoprint instances both ways, that all are up to date and in case you run jobs from an octoprint directly, the correct status of the Spools are used by selecting the spool on the actual used Octoprint.

### LINKED ISSUE #353: Filament Manager plugin is not ignoring the Filter's on the Spools page when displaying the information in the overview table.

NotExpectedYet:
Major clean up of filament manager. 

To Do: 
   - [x] Fixed #346 
   - [x]  Fixed #353 
   - [x] Clean up of the ejs template loading values not disabled
   - [x] Enable values for editing, then disable again on save
   - [x] Fix issue with overview table not updating fields when filter applied to spool manager
   - [x] Fix issue with overview table not displaying correct values
   - [x] Added conditional turnary operator to big please @davidzwa "?." 
   - [x] 20 Squats to appease the printing gods 
   - [x] 10 pushups to appease big D
   - [x] Fix adding a spool would not add disabled
   - [x] Fix innerHTML of undefined when adding new spool
   - [x] Fix NaN values on total price / weight on Spool Manager


davidzwa (approved):
LGTM, as far as I can see. Havent tested it locally, but I trust youve worked hard enough making this robust.

PR #391, opened 2021-03-24
Bugfix/update docker sh scripts

NotExpectedYet:
Docker scripts need to use npm ci --production going forward like start script.
 - [x] Fix docker*.sh scripts
 - [x] Fix monolithic dockerfile as not in script
 - [x] 5 squats
 - [x] 5 push ups


NotExpectedYet:
grrr, sorry forgot to rebase off development. Not an issue no?

davidzwa (commented):
Cant approve yet:
`npm install --only=production` does not clear out the modules where `npm ci --production` does do this. 

We can have deletion/file lock errors if we do this while the pm2 nodejs octofarm service still has lock on modules. That's why I left it at npm install --production earlier.

Must admit these are lessons learned on my side as well!

NotExpectedYet:
> Cant approve yet:
> `npm install --only=production` does not clear out the modules where `npm ci --production` does do this.
> 
> We can have deletion/file lock errors if we do this while the pm2 nodejs octofarm service still has lock on modules. That's why I left it at npm install --production earlier.
> 
> Must admit these are lessons learned on my side as well!

I gather that's only applicable to the monolithic branch as the other two check for the existence of the node_modules folder whereas monolithic doesn't. 

I'll revert that one if that's the case.

davidzwa:
I think I mustve been asleep still, now that I think of it .... node_modules being locked or not does not matter at all... it's an entrypoint script 

Ugh I feel so dumb xD

NotExpectedYet:
> I think I mustve been asleep still, now that I think of it .... node_modules being locked or not does not matter at all... it's an entrypoint script ... so nothing can be using the modules at start.
> 
> Ugh I feel so dumb xD

The entry point scripts aye, but is that still the case for the RUN commands? I gather they don't get re-run on a docker restart. A rebuild wouldn't matter as they'd be no cache of that folder.

davidzwa:
RUN = build time only. That's done on github :P

PR #392, opened 2021-03-24
Bugfix/incorrect print progress

NotExpectedYet:
I have checked all the code from grabbing from OctoPrint -> Displaying on OctoFarm and all seems fine. 

There is a large discrepency with the print time remaining which I believe caused this issue. 

Tested anyway and follows in line now with the correct progress to the end of a completed print so I believe the issue is resolved. 

 - [x] Use OctoPrints print time remaining + elapsed print time to generate OctoFarms expected print time value. More closely inline with OctoPrints although that doesn't seem to keep the value updated at all after a print is started. I believe OctoFarm should supersede that calculation as mine is more to real life.  Example below. OctoFarm was inline with Octoprint until PrintTimeLeft value updated, OctoPrints UI didn't update this. 
 - [x] Make OctoFarm round down percentage as this now displays 100% when infact there is still a slight bit of printing to do. 
 
 ##Print time remaining dependency. 
image
image

##Progress inline - At worst this get's out by a single percent due to OctoFarm rounding up, and OctoPrint rounding down. Updated above to change this to round down. 
image
image


davidzwa:
@NotExpectedYet is the second todo done as well?

davidzwa (approved):
LGTM and I trust you tested.

davidzwa (approved):

NotExpectedYet:
sweet I'll pull locally and sort the last check box shortly. Cheers!

NotExpectedYet:
@davidzwa ready :)

davidzwa (approved):

PR #395, opened 2021-03-24
Feature/packaging and app fixes

davidzwa:
Ok here goes:
- [x] improve database check at startup
- [x] fix views folder (setup step was skipped, or I presume left default causing all the views errors in the past)
- [x] add package command
- [x] fix and test copyFile command, failed locally
- [x] text .env file working
- [x] test .exe working
- [x] test image default copied
- [x] test logs fine
- [x] add app-pancaked for pkg env fixes
- [x] add redirect * for both database-issues and normal mode (both tested)
- [x] fix printerclean environmentalData undefined error in case of database issues
- [ ] (TODO) fix printerclean heatmap [0] undefined error in case of database issues

We should consider aborting if the MONGO variable is not set, or add smart EJS+UI to configure it. The latter? This PR or the next?
We should in general not load any runners... that would fix stuff altogether...


NotExpectedYet:
I like the UI to configure it certainly! If it isn't local it's gonna be the easiest way for people to configure these options. 

Man nice work just checking through this.

NotExpectedYet:
If you wanna handle it this feel free, doesn't really make a difference to me. There's the database page you can use to create it. Or I can do it Thursday in another PR totally fine.

davidzwa:
@NotExpectedYet fire away!

NotExpectedYet (approved):
Sweet as!

davidzwa:
Are you willing to test the docker containers? Maybe script that?
- build image for the 3 types
- create 3 containers for each type
- boot them on a different api port each
- call them on dashboard page

NotExpectedYet:
I need to check all tbh, the packages as well as because I wanna make sure history is working which it wasn't in my tests yesterday.

davidzwa:
Lets bash script it then!

PR #406, opened 2021-03-27
Feature/366 analyze and cleanup history

davidzwa:
Going strong on cleanup and full unit tests (never did it this thoroughly before).
Replacing code only by full undefined/empty/null/falsy/truthy tests, which is massively valuable so far.

In `historyClean.js`, `fileClean.js` and `printerClean.js`:
- `checkNestedIndex(...)`, `checkNested(...)` broken out and revamped
- `getPrintCost(...)` broken out and tested slightly for NaN (it wasnt adjusted)
-

### DIFF
--- a/server_src/lib/tests/history-clean.test.js
+++ b/server_src/lib/tests/history-clean.test.js
@@ -1,0 +1,31 @@
+jest.mock("../../services/history.service");
+const mockHistoryService = require("../../services/history.service");
+const { HistoryClean } = require("../dataFunctions/historyClean");
+
+describe("historyClean", function () {
+  it("should start and finish within 5 sec for empty history", async function () {
+    expect(await mockHistoryService.find({})).toHaveLength(0);
+
+    const historySyncService = new HistoryClean(false, "info");
+    expect(historySyncService.logger).toBeDefined();
+    const voidResponse = await historySyncService.start();
+    expect(voidResponse).toBeFalsy();
+    expect(historySyncService.statisticsClean).toBeTruthy();
+    expect(historySyncService.historyClean).toBeTruthy();
+    expect(historySyncService.historyClean).toHaveLength(0);
+    expect(true).toEqual(true);
+  });
+
+  // it("should start and finish within 5 sec for empty history", async function () {
+  //   expect(await mockHistoryService.find({})).toHaveLength(0);
+  //
+  //   const historySyncService = new HistoryClean(false, "info");
+  //   expect(historySyncService.logger).toBeDefined();
+  //   const voidResponse = await historySyncService.start();
+  //   expect(voidResponse).toBeFalsy();
+  //   expect(historySyncService.statisticsClean).toBeTruthy();
+  //   expect(historySyncService.historyClean).toBeTruthy();
+  //   expect(historySyncService.historyClean).toHaveLength(0);
+  //   expect(true).toEqual(true);
+  // });
+});
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,6 +12,7 @@ module.exports = {
     parser: "babel-eslint",
     sourceType: "module",
   },
+  ignorePatterns: [".eslintrc.js"],
   rules: {
     "prettier/prettier": "error",
     "max-len": [
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,5 @@
+{
+  "singleQuote": false,
+  "trailingComma": "all",
+  "tabWidth": 2
+}
--- a/client_src/js/lib/functions/calc.js
+++ b/client_src/js/lib/functions/calc.js
@@ -63,6 +63,7 @@ export default class Calculate {
     return (n - a) * (n - b) <= 0;
   }
 
+  // TODO EXACT COPY FROM SERVER print-cost.util.js!!
   static returnPrintCost(costSettings, time) {
     if (typeof costSettings === "undefined") {
       // Attempt to update cost settings in history...
--- a/server_src/lib/dataFunctions/fileClean.js
+++ b/server_src/lib/dataFunctions/fileClean.js
@@ -1,5 +1,6 @@
 const _ = require("lodash");
 const Logger = require("../logger.js");
+const { getPrintCost } = require("../utils/print-cost.util");
 
 const logger = new Logger("OctoFarm-InformationCleaning");
 
@@ -131,7 +132,7 @@ class FileClean {
               failed: file.failed,
               last: file.last,
               expectedPrintTime: file.time,
-              printCost: await FileClean.getPrintCost(file.time, printCost),
+              printCost: getPrintCost(file.time, printCost),
             };
             sortedFile.toolUnits = await FileClean.getUnits(
               selectedFilament,
@@ -221,29 +222,6 @@ class FileClean {
       return strings;
     }
     return [];
-  }
-
-  static async getPrintCost(printTime, costSettings) {
-    if (typeof costSettings === "undefined") {
-      // Attempt to update cost settings in history...
-      return "No cost settings to calculate from";
-    }
-    // calculating electricity cost
-    const powerConsumption = parseFloat(costSettings.powerConsumption);
-    const costOfElectricity = parseFloat(costSettings.electricityCosts);
-    const costPerHour = powerConsumption * costOfElectricity;
-    const estimatedPrintTime = printTime / 3600; // h
-    const electricityCost = costPerHour * estimatedPrintTime;
-    // calculating printer cost
-    const purchasePrice = parseFloat(costSettings.purchasePrice);
-    const lifespan = parseFloat(costSettings.estimateLifespan);
-    const depreciationPerHour = lifespan > 0 ? purchasePrice / lifespan : 0;
-    const maintenancePerHour = parseFloat(costSettings.maintenanceCosts);
-    const printerCost =
-      (depreciationPerHour + maintenancePerHour) * estimatedPrintTime;
-    // assembling string
-    const estimatedCost = electricityCost + printerCost;
-    return estimatedCost.toFixed(2);
   }
 
   static async getCost(filamentSelection, units) {
--- a/server_src/lib/dataFunctions/historyClean.js
+++ b/server_src/lib/dataFunctions/historyClean.js
@@ -1,7 +1,10 @@
-const ServerSettings = require("../../models/ServerSettings.js");
-const Logger = require("../logger.js");
+const { checkNested, checkNestedIndex } = require("../utils/array.util");
+const { getPrintCost } = require("../utils/print-cost.util");
 const { getHistoryCleanDefault } = require("../providers/history.constants");
+const ServerSettings = require("../../models/ServerSettings.js");
 const historyService = require("../../services/history.service");
+
+const Logger = require("../logger.js");
 
 class HistoryClean {
   enableLogging = false;
@@ -17,30 +20,6 @@ class HistoryClean {
       logLevel,
     );
     this.enableLogging = enableFileLogging || false;
-  }
-
-  static checkNested(nameKey, myArray) {
-    try {
-      for (var i = 0; i < myArray.length; i++) {
-        if (myArray[i].name === nameKey) {
-          return myArray[i];
-        }
-      }
-    } catch (e) {
-      logger.error("Couldn't check nested....", JSON.stringify(e));
-    }
-  }
-
-  static checkNestedIndex(nameKey, myArray) {
-    try {
-      for (var i = 0; i < myArray.length; i++) {
-        if (myArray[i].name === nameKey) {
-          return i;
-        }
-      }
-    } catch (e) {
-      logger.error("Couldn't check nested index...", JSON.stringify(e));
-    }
   }
 
   static async getHours(printTime) {
@@ -100,6 +79,7 @@ class HistoryClean {
       }
     }
 
+    // TODO remove as it is simply a nullable call
     //Get spoolid function
     function spoolID(id) {
       if (typeof id !== "undefined" && id !== null) {
@@ -251,30 +231,6 @@ class HistoryClean {
     }
   }
 
-  static getPrintCost(printTime, costSettings) {
-    if (typeof costSettings === "undefined") {
-      //Attempt to update cost settings in history...
-      return "No cost settings to calculate from";
-    } else {
-      // calculating electricity cost
-      const powerConsumption = parseFloat(costSettings.powerConsumption);
-      const costOfElectricity = parseFloat(costSettings.electricityCosts);
-      const costPerHour = powerConsumption * costOfElectricity;
-      const estimatedPrintTime = printTime / 3600; // h
-      const electricityCost = costPerHour * estimatedPrintTime;
-      // calculating printer cost
-      const purchasePrice = parseFloat(costSettings.purchasePrice);
-      const lifespan = parseFloat(costSettings.estimateLifespan);
-      const depreciationPerHour = lifespan > 0 ? purchasePrice / lifespan : 0;
-      const maintenancePerHour = parseFloat(costSettings.maintenanceCosts);
-      const printerCost =
-        (depreciationPerHour + maintenancePerHour) * estimatedPrintTime;
-      // assembling string
-      const estimatedCost = electricityCost + printerCost;
-      return estimatedCost.toFixed(2);
-    }
-  }
-
   static getFile(history) {
     const file = {
       name: history.fileName,
@@ -371,7 +327,7 @@ class HistoryClean {
               const keys = Object.keys(spool);
               for (const key of keys) {
                 //check if type exists
-                let checkNested = this.checkNested(
+                let checkNested = checkNested(
                   JSON.parse(JSON.stringify(spool[key].type)),
                   totalByDay,
                 );
@@ -379,17 +335,17 @@ class HistoryClean {
                 if (typeof checkNested !== "undefined") {
                   let checkNestedIndexHistoryRates = null;
                   if (historyClean[h].state.includes("success")) {
-                    checkNestedIndexHistoryRates = this.checkNestedIndex(
+                    checkNestedIndexHistoryRates = checkNestedIndex(
                       "Success",
                       historyByDay,
                     );
                   } else if (historyClean[h].state.includes("warning")) {
-                    checkNestedIndexHistoryRates = this.checkNestedIndex(
+                    checkNestedIndexHistoryRates = checkNestedIndex(
                       "Cancelled",
                       historyByDay,
                     );
                   } else if (historyClean[h].state.includes("danger")) {
-                    checkNestedIndexHistoryRates = this.checkNestedIndex(
+                    checkNestedIndexHistoryRates = checkNestedIndex(
                       "Failed",
                       historyByDay,
                     );
@@ -397,7 +353,7 @@ class HistoryClean {
                     return;
                   }
 
-                  let checkNestedIndexByDay = this.checkNestedIndex(
+                  let checkNestedIndexByDay = checkNestedIndex(
                     JSON.parse(JSON.stringify(spool[key].type)),
                     usageOverTime,
                   );
@@ -418,7 +374,7 @@ class HistoryClean {
                     );
                   }
 
-                  let checkNestedIndex = this.checkNestedIndex(
+                  let checkNestedIndex = checkNestedIndex(
                     JSON.parse(JSON.stringify(spool[key].type)),
                     totalByDay,
                   );
@@ -467,8 +423,6 @@ class HistoryClean {
                         x: dateParse,
                         y: weightCalcSan,
                       });
-                      // console.log(checkNestedIndexHistoryRates);
-                      // console.log(historyByDay[checkNestedIndexHistoryRates].name);
                       historyByDay[checkNestedIndexHistoryRates].data.push({
                         x: dateParse,
                         y: 1,
@@ -682,7 +636,7 @@ class HistoryClean {
           endDate: currHistory.endDate,
           printTime: currHistory.printTime,
           notes: currHistory.notes,
-          printerCost: HistoryClean.getPrintCost(
+          printerCost: getPrintCost(
             currHistory.printTime,
             currHistory.costSettings,
           ),
--- a/server_src/lib/dataFunctions/historyClean.js
+++ b/server_src/lib/dataFunctions/historyClean.js
@@ -1,135 +1,22 @@
 const ServerSettings = require("../../models/ServerSettings.js");
-const History = require("../../models/History.js");
 const Logger = require("../logger.js");
-const PrinterClean = require("../dataFunctions/printerClean.js");
-
-const logger = new Logger("OctoFarm-InformationCleaning");
-let historyClean = [];
-let statisticsClean = {
-  completed: 0,
-  cancelled: 0,
-  failed: 0,
-  longestPrintTime: 0,
-  shortestPrintTime: 0,
-  averagePrintTime: 0,
-  mostPrintedFile: 0,
-  printerMost: 0,
-  printerLoad: 0,
-  totalFilamentUsage: 0,
-  averageFilamentUsage: 0,
-  highestFilamentUsage: 0,
-  lowestFilamentUsage: 0,
-  totalSpoolCost: 0,
-  highestSpoolCost: 0,
-  totalPrinterCost: 0,
-  highestPrinterCost: 0,
-  currentFailed: 0,
-};
+const { getHistoryCleanDefault } = require("../providers/history.constants");
+const historyService = require("../../services/history.service");
 
 class HistoryClean {
-  static returnHistory() {
-    return historyClean;
-  }
+  enableLogging = false;
+  historyClean = [];
+  logger;
+  statisticsClean = getHistoryCleanDefault();
+  historyService = historyService;
 
-  static returnStatistics() {
-    return statisticsClean;
-  }
-
-  static async start() {
-    logger.info("Running history Cleaning");
-    try {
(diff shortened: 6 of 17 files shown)


davidzwa:
This pull-request will eventually trample like a stampede over every other branch which changes runners, lib or routes. This is not permanent, but required as I was crunching code in historyClean which was tighly coupled to printerClean, printer.js, history.js and some other dataFunctions.

Looking at the amount of changes, I'd propose to review and run the tests (and not the code). These tests were hand-crafted by me to convey the message and show how I was thinking/my intent.
After that we can talk code, but only after that.

NotExpectedYet:
No worries, I'll take a look Wednesday anyway as working till then.

davidzwa:
It's time.

NotExpectedYet:
Sweet, not too far off history so I'll get that finished and take a look.

NotExpectedYet:
Doesn't run as JobClean isn't updated with the changes 

`const { getPrintCostNumeric } = require("../utils/print-cost.util");` - Missing
`      const filamentEntry = Array.isArray(filamentSelection)
        ? filamentSelection[m]
        : filamentSelection;` - This line produces a `m` is not defined error, I gather due to the for in loop rather than my for loop as I use the index of `m`

NotExpectedYet (changes requested):
Doesn't run as JobClean isn't updated with the changes 

`const { getPrintCostNumeric } = require("../utils/print-cost.util");` - Missing and not updated
`      const filamentEntry = Array.isArray(filamentSelection)
        ? filamentSelection[m]
        : filamentSelection;` - This line produces a `m` is not defined error, I gather due to the for in loop rather than my for loop as I use the index of `m`

This is on my database which have printers and spools attached to each.

server_src/cache/printer.cache.js
```diff
@@ -0,0 +1,36 @@
...
+
+function returnCurrentOperations() {
+  return printerCleanState.currentOperations;
+}
+
+function returnDashboardStatistics() {
+  return printerCleanState.dashboardStatistics;
+}
+
+module.exports = {
```
NotExpectedYet, line 29: Is there a reason for grouping all the exports here? 

I've opted for exporting at the function level. If it's a preference rather than anything functional then I'm happy to switch mine out and do as an export at the bottom. Just want to keep these consistent.

server_src/lib/providers/settings.constants.js
```diff
@@ -0,0 +1,53 @@
+function getDefaultLayout() {
```
NotExpectedYet, line 1: Is it worth keeping these inside of lib? I personal think they would be better all under the services folder but let me know your thoughts.

NotExpectedYet (changes requested):

server_src/cache/printer.cache.js
davidzwa: I thought about it a bit, when we switch to Typescript all these functions will go into a class and that class will be prepended with `export`. So having to delete one block at the bottom with module.exports is easiest.
Also, I find the explicit block at the bottom much cleaner.

server_src/lib/providers/settings.constants.js
davidzwa: These are app constants. I am collecting them to distribute them into V2 modules later, so they would end up in a `blablab.config.ts` or `blablabla.constants.ts` in the root or config folder of a module. If the constant is really important, it would go to `app.constants.ts`.

So this is me preparing for that.

davidzwa:
> Doesn't run as JobClean isn't updated with the changes
> 
> `const { getPrintCostNumeric } = require("../utils/print-cost.util");` - Missing and not updated
> ` const filamentEntry = Array.isArray(filamentSelection) ? filamentSelection[m] : filamentSelection;` - This line produces a `m` is not defined error, I gather due to the for in loop rather than my for loop as I use the index of `m`
> 
> This is on my database which have printers and spools attached to each.

@NotExpectedYet this is a good catch and exactly what I needed. Many thanks. Now my question is, is it ok that I removed the array flattening action? What would the impact be?

server_src/cache/printer.cache.js
NotExpectedYet: Sound I'll switch mine out to a single export at the bottom.

server_src/lib/providers/settings.constants.js
NotExpectedYet: Good to know! Cheers.

NotExpectedYet:
> > Doesn't run as JobClean isn't updated with the changes
> > `const { getPrintCostNumeric } = require("../utils/print-cost.util");` - Missing and not updated
> > ` const filamentEntry = Array.isArray(filamentSelection) ? filamentSelection[m] : filamentSelection;` - This line produces a `m` is not defined error, I gather due to the for in loop rather than my for loop as I use the index of `m`
> > This is on my database which have printers and spools attached to each.
> 
> @NotExpectedYet this is a good catch and exactly what I needed. Many thanks. Now my question is, is it ok that I removed the array flattening action? What would the impact be?

I'm cycling through the filament selected on tools here which are stored in an array. The index of where it is in the array is used to define the tool number as well. We'd loose the calculations on the specific tools if I'm understanding you correctly. So Tool 0: printed 20g, Tool 1: printed 55g. Here it's mainly used to generate the History card pop-up where the spool table is. It wouldn't effect any totals just stop a breakdown.

davidzwa:
### Questions
1) Let me cut straight to the chase, why is the input for HistoryClean.getSpool(filamentSelection) not always an array? Looking at that Array.isArray check I found there earlier.

2) Second question - is 'selection' in the 'filamentSelection' variable name indicating that it will be an array? To me selection can indicate a single element as well as array haha. If I understand your word usage, it might guide me into the right direction for other file refactorings as well (to understand your naming/wording a bit better and understand the context of the code).

3) Third question, I see that HistoryClean is being called a lot by other cleaner files, as well as our friend `state.js` which I see as the root of the tree. But other cleaners have similar functions like getSpool and getSpools, just to give 1 example. Are those functions doing the same or are they contextually different? Can we untangle that a bit or find better names for them? (Or maybe there is code duplication?)

4) Fourth, are you ok with removing async/awaits where they are not used on any code you will be tackling soon? Async/awaits to me indicate that calling that code 'might' introduce delays and I would then avoid using such functions unless I need to. Then after looking at the functions, I sometimes notice they are not async at all and suddenly the whole code structure becomes so much simpler (no awaits, `then(...)` and `.catch(...)` everywhere).

5) Can we remove `checkNestedIndex` and `checkNested`? I dont understand the data structure and because there's no Typescript models I have no clue what the code is doing. The only thing I see is magical Object.keys(...) for-looping which I intensely hope can be avoided. Your insight counts heavily here.

### Analysis
Some big things - could be called code smells - I noticed that we can talk about. Some are slightly subjective, but I wont hold back.
- DONT define functions within functions. I didnt even saw you did until I reached that code.... I scrolled over it many many times (I collapse big functions as that is the only way for me to keep overview) 🟥
- 1 function, 1 responsibility (Example: if we generate statistics, we should not query serverSettings. Someone should provide that to us as argument - in this case `filamentManagerEnabled: boolean` or the relevant combination of statistics related settings, not more not less) 🟥
- always catch promises at the highest possible level (route or runner level). Same thing, catch as late as possible so you have the information required. This is not a rule for all situations, ofc sometimes you have exceptions down below. But make sure you can still 'humanize' the exceptions. Too deep and you simply dont have enough information to handle the error or the reason why the exception happened 🟥
- dont try catch everything. Exceptions exist for a reason. Catch as late as possible and as little as possible. (catching exceptions hide the problem, instead we should find the problem asap with all details/context required to resolve it) 🟥
- reduced indentation of code. More than 4 indentations and most coders start to wonder where they got lost in the rabbit hole. Use functions and de-duplicate any functions once they start to do similar things. This is the easiest rule to create cleaner code. 🟥
- JSON.parse(JSON.stringify()) on elements which are not arrays or objects. Even then it's [a really bad idea for f.e. Date objects](https://stackoverflow.com/questions/122102/what-is-the-most-efficient-way-to-deep-clone-an-object-in-javascript). Deep cloning without proper argumentation is a big red flag as it does not answer the question: 'why are you cloning a heavily nested object and why are you using a method which is advised against concerning data loss?'. If you need to shallow copy, we have the spread operator with Node 8+, much faster and does not trigger a red flag for devs - TypeScript will really help out with the resulting type as well. 🟥
- 

- removed duplicate code and constants. Dont repeat yourself. Ever. 🟧 
- lot of array accessors being repeated, removed most of them. I dont get why it would be of advantage. Do explain, because it's all over the code. 🟧 
- lot of (often deeeeeeply nested) property being repeatedly accessed, makes it really hard to predict at which level of object nesting we are (this requires an example on my side probably) 🟧 
- deeeeeeeply nested database objects. Huge source of bugs. I don't hate nesting, but there are cases where I think it's pushed to the extreme and totally unnecessary. Spools are one case. 🟧
- property wrong or null? return immediately (avoid long `if {...}` followed by really short `else {...}`, just a `if(!....) return null;` and the validation part of code becomes really much more readable/at the start of each function) 🟧
- removal of 'typeof undefined' - as discussed 🟨 
- TODO we should add a global exception handler to ExpressJS/NodeJS 🟩

NotExpectedYet:
> ### Questions
> 
>     1. Let me cut straight to the chase, why is the input for HistoryClean.getSpool(filamentSelection) not always an
> array? Looking at that Array.isArray check I found there earlier.

    That's legacy from the transition to multi-tool stuff. I don't think we can remove because we might have some people with long standing history where the filamentSelection is captured as a single object, not an array of Objects. Old history before the multi-tool selection would have the single object, whereas newer history will have the Array. We'd at the very least need to check and convert to an Array maybe at the database level. Then it wouldn't be so much of an issue here. 

> 
>     2. Second question - is 'selection' in the 'filamentSelection' variable name indicating that it will be an array? To me selection can indicate a single element as well as array haha. If I understand your word usage, it might guide me into the right direction for other file refactorings as well (to understand your naming/wording a bit better and understand the context of the code).
>     
    filamentSelection is defining the currently selected spool from the dropdown in the UI that's then captured in the history.js runner, as in This Filament/Spool was selected, naming could be more explicity I agree. As above, it should be an Array but depends how old a users history was. 
> 
>     3. Third question, I see that HistoryClean is being called a lot by other cleaner files, as well as our friend `state.js` which I see as the root of the tree. But other cleaners have similar functions like getSpool and getSpools, just to give 1 example. Are those functions doing the same or are they contextually different? Can we untangle that a bit or find better names for them? (Or maybe there is code duplication?)
>     
      Code smells, needs cleaning up with a single source of truth type system. The problem was I saw the utility in grabbing some of the cleaned information in those files. The similar functions was stoner laziness. That can be untangled totally and given better names. There will certainly be code duplication as the HistoryClean and JobClean user similar functions as we found out with the recent crash hehe. 
> 
>     4. Fourth, are you ok with removing async/awaits where they are not used on any code you will be tackling soon? Async/awaits to me indicate that calling that code 'might' introduce delays and I would then avoid using such functions unless I need to. Then after looking at the functions, I sometimes notice they are not async at all and suddenly the whole code structure becomes so much simpler (no awaits, `then(...)` and `.catch(...)` everywhere).
>     
     If they still work the same yes that's fine. I had a lot of async and await stuff because I experienced client slow downs when making system calls and couldn't find the true cause and I tried that on the offchance and it fixed it. I think the issue would be with the interval stuff in that regard. I thought they we're only syntactic sugar personally so if there's an underlying difference be happy to hear it as I prefer reading await > .then and you can get into some messy situations with .thens. Saying that I've not entirely followed that myself everywhere XD. 
> 
>     5. Can we remove `checkNestedIndex` and `checkNested`? I dont understand the data structure and because there's no Typescript models I have no clue what the code is doing. The only thing I see is magical Object.keys(...) for-looping which I intensely hope can be avoided. Your insight counts heavily here.
>     
      I don't think so personally, the checkNestedIndex and checkNested search for the keys when building the graphing arrays up. So I check if the index exists for example, then if it doesn't I create the array there and then. The issue was with the ones surrounding spools and such, I have no idea how many tools some users might have on their printers so seemingly at the time I had to at least check for the existence first. I'd have to double check which functions are using it and have a look again at building the arrays as there could be a better way totally. Excuse my ignorance but what's wrong with Object.keys? 
> 
>     6. Couple of TODO's in the static method `HistoryClean.getSpool(...)`. Why the assignment `metrics = metrics.filament;`? Also, fun story, it was a really cool puzzle to see the whole getSpool function collapse when I applied some tricks xD It took me like multiple hours of staring and everytime I saw something different to strip off and suddenly almost nothing was left hahahah
>        Btw, is this acceptable and what we want? No better fix?
>        
         I assign metrics just to break out of the object, as all I want is the filament object. I gather that's more of an issue with how the spool is passed through above, I can't rightly remember right now. The code is older than it looks on git btw because I'd copied a lot of those functions through from when they we're in the UI. Some of it is specifically in there because of how the old databases might be that the user has. That's a mixed bag too. 
         
         Yeah you've opened my can of worms with that one XD. It's a system of tangled web's built from hemp ;) and smokey afternoon epiphany's on the can. It needs decoupling first in my opinion then we can look at simplifying the systems and data structures, but happy for more input like this thanks! 
> 
> ```
> // Fix for old database states
>    if (!metrics?.filament) {
>       return null;
>     }
> ```
> 
 I don't believe there's anytihng wrong with that, off top of my head anyway. We require the filament object for the functions below regardless. So that was just stopping them generating. It's much simpler now we have ? operators hehe. 
> ### Analysis
> 
This is why we cleaning :P

NotExpectedYet:
@davidzwa 

`(node:7862) UnhandledPromiseRejectionWarning: TypeError: HistoryClean.returnHistory is not a function
    at /home/s1mpleman/Development/NotExpectedYet Development/Testing/OctoFarm/server_src/routes/index.js:155:38
    at Layer.handle [as handle_request] (/home/s1mpleman/Development/NotExpectedYet Development/Testing/OctoFarm/node_modules/express/lib/router/layer.js:95:5)`
    
This is when loading the history page

NotExpectedYet:
`Firefox can’t establish a connection to the server at http://localhost:4000/dashboardInfo/get/.`

Get this client side too when loading any page. Pops up and goes away.

davidzwa:
- I get "Uncaught (in promise) TypeError: Cannot read property 'farmActivity' of undefined" on dash
- You got "dashboard is undefined" on dash
- And you got "HistoryClean.returnHistory is not a function" in server index.js

Back to work!

davidzwa:
@NotExpectedYet it should be fixed in the rebased branch! Some functions in index.js werent correctly calling the new cache

NotExpectedYet:
`KEY tool0`
`METRICS { tool0: { volume: 4.75537618508613, length: 672.7481121048331 } }`

`(node:660364) UnhandledPromiseRejectionWarning: TypeError: metrics.indexOf is not a function
    at Function.getSpool (/home/s1mpleman/Development/NotExpectedYet Development/Testing/HistoryClean/server_src/lib/dataFunctions/historyClean.js:149:32)
    at HistoryClean.initCache (/home/s1mpleman/Development/NotExpectedYet Development/Testing/HistoryClean/server_src/lib/dataFunctions/historyClean.js:464:30)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
(node:660364) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
`

New error after getting it to run.

davidzwa:
The indexof bug was a fix never tested by me. Awesome catch!
I have moved cache initiation to the end of index.js. The reason is that I want routes to be loaded before we start doing CPU heavy work. Maybe we can do this more? This way we can mark the API as loading just like OctoPrint does "loading OctoPrint...."

NotExpectedYet:
> The indexof bug was a fix never tested by me. Awesome catch!
> I have moved cache initiation to the end of index.js. The reason is that I want routes to be loaded before we start doing CPU heavy work. Maybe we can do this more? This way we can mark the API as loading just like OctoPrint does "loading OctoPrint...."

No that's fine, we should control it a lot better going further I agree.

davidzwa:
Without disclosing unreviewed but obfuscated history test data, I can share that the calculatest are super robust so far.
The only side note is the following:
```
      "totalSpoolCost": "NaN", // TODO interesting
      "highestSpoolCost": "NaN", // TODO interesting
```

Will provide the test functionality later when the test set is approved for public eyes.

NotExpectedYet:
> ### Testing results
> 
> Without disclosing unreviewed but obfuscated history test data, I can share that the calculations are super robust so far **on the domain level**. We still have to provide more **regression tests** to provide more contextual validation of the cleaner functionality (we are not testing cost calculations fully yet) and API tests.
> 
> The stats of a local test set of 14 realistic history entries with mixed properties result in the following interesting properties:
> 
> ```
>       "totalSpoolCost": "NaN", // TODO interesting
>       "highestSpoolCost": "NaN", // TODO interesting
> ```
> 
> **The conclusion from the test set is the following:**
> The cost calculation fails to skip entries which do not have a spool. We should decide: do we bail out or filter out History entries without tool specification? An additional option is to filter out those entries and add warnings to the calculation results, like f.e. "The history costs might be inaccurate because {0} jobs are missing spool definitions".
> _Will provide the tests and mock data later when the test set is approved for public eyes._

I like the second options better for user interaction but I see no dire need for it. Could be an upgrade for later, the calculation should be defaulting to PLA values. A note could be worth adding regardless on in a "title" tag on the html element as these are not much explained.

NotExpectedYet (changes requested):
Few issues: 
Total cost field displays NaN with no spool. Should revert to Printer Cost only in calculation. 
Total Spool Cost and Highest Spool Cost display NaN, these should be ignoring NaN values. 
Print Completion by day graph doesn't load.
Non of the Filament Usage graphs load. 

This isn't ready for prime time yet ma dude, but epic work so far everything else generated or left blank for spools which is good as it deals with the older database structure and the new for dual/single spools. 

I believe the graphs for filament might be due to the possibility of the job field been "null"

NotExpectedYet:
> Few issues:
> Total cost field displays NaN with no spool. Should revert to Printer Cost only in calculation.
> Total Spool Cost and Highest Spool Cost display NaN, these should be ignoring NaN values.
> Print Completion by day graph doesn't load.
> Non of the Filament Usage graphs load.
> 
> This isn't ready for prime time yet ma dude, but epic work so far everything else generated or left blank for spools which is good as it deals with the older database structure and the new for dual/single spools.
> 
> I believe the graphs for filament might be due to the possibility of the job field been "null"

I'm wondering if the graphs didn't load because I didn't have any data in the last 30 days. I've got it on the development branch so I'd take a guess that's correct.

davidzwa:
> I'm wondering if the graphs didn't load because I didn't have any data in the last 30 days. I've got it on the development 
branch so I'd take a guess that's correct.

Only thing I can say, and Im completely guessing here, is that `undefined` doesnt get sent by javascript. That's a `fetch` default I guess. If something is not set why send it?

Now gonna fix the cost NaN's.

NotExpectedYet:
> Only thing I can say, and Im completely guessing here, is that `undefined` doesnt get sent by javascript. That's a `fetch` default I guess. If something is not set why send it? I changed some nulls to undefined, and that might need proper exchange of thoughts as that is a choice not a preference.
> 
> Now gonna fix the cost NaN's.

I'm not sure how undefined would come into play here as apexCharts will just look at an array with spoolType: "PLA", data: [] for example. So the data fields would never contain any actual data. There's a simple enough test for it and that's to run a few prints off and see if they generate.

server_src/lib/tests/history-clean.test.js
```diff
@@ -0,0 +1,277 @@
...
+      "totalPrinterCost": "7.62",
+      "highestPrinterCost": "1.89",
+      "currentFailed": 247,
+      "totalByDay": [],
+      "usageOverTime": [],
+      "historyByDay": []
+    });
+
+    expect(stats.totalSpoolCost).not.toBe("NaN");
+    expect(stats.highestSpoolCost).not.toBe("NaN");
```
davidzwa, line 136: So we now check that it is not string NaN.

davidzwa:
> I'm not sure how undefined would come into play here as apexCharts will just look at an array with spoolType: "PLA", data: [] for example. 

Ok then its the "NaN" strings. Apex hates those, let alone NaN.

> So the data fields would never contain any actual data. There's a simple enough test for it and that's to run a few prints off and see if they generate.

Please spec this in written words to the minute detail! This I can write a test for and we dont need OP or real life at first. Fail quick, fail fast.

davidzwa:
image
Proof

davidzwa:
image
Proof

NotExpectedYet:
> > I'm not sure how undefined would come into play here as apexCharts will just look at an array with spoolType: "PLA", data: [] for example.
> 
> Ok then its the "NaN" strings. Apex hates those, let alone NaN.
> 
> > So the data fields would never contain any actual data. There's a simple enough test for it and that's to run a few prints off and see if they generate.
> 
> Please spec this in written words to the minute detail! This I can write a test for and we dont need OP or real life at first. Fail quick, fail fast.

You'd need max 3 prints copied from history but with their state and end dates generated in the last 30 days for the graphs to generate. 

They look a bit weird with 1 or 2 values.

NotExpectedYet:
> image
> Proof

There was fine xD sorry. I meant the table rows for the individual totals.

NotExpectedYet (changes requested):
This is coming along awesomely man. 

image
1. The spools are sometimes null here. Could these generate the (No Spool) tag for a tool 0 at all? Like a fall back. It's something I setup to allow a user to go update down the road if they wish and the blank field breaks opening the history card. 

I'm still getting nothing in the graph data btw. historyByDay, usageOverTime, totalByDay are all coming back as empty arrays. These are all recent prints with spools attached, and amounts used. 

The statistic button on printer statistics doesn't load anymore either. 
history/statistics/:id

Gather it just needs updating to the cache.

That's pretty much all, then might as well get this in 1.1.14. Again nice work.

davidzwa:
As discussed offline I'll work on fixing the history collection before we use the data in history clean.

davidzwa:
1) I have checked develop and the tool not showing up is a thing that happens there as well. This is code I kept intact, although I did have multiple questions why. Your code explicitly returns `null` for metrics being undefined (old database)
https://github.com/OctoFarm/OctoFarm/blob/e99cabd3f27f3ab6e21c27f0da5e8e319c4a2751/server_src/lib/dataFunctions/historyClean.js#L704
2) The getSpool() function still returns an array which is saved to `spools`, nothing wrong there. So point 1 applies.
3) The graph and statistics button ... gotta look into that

Here's dev:
image

davidzwa:
**30 or 90 days ago** - @NotExpectedYet I found the reason why the graph was empty. Your variable thirtyDaysAgo was actually assigned a date of 90 days ago... so yeah my code was actually doing 30 days ago. Im at least I glad I found it.

**Loading** - About the loading... I made #549 and you made #538. So I think that is fixed in another PR.

**Calculation** Final thing to note is that my code is outputting **higher usage numbers**. Let me show you:
Previous code:
```
    "totalFilamentUsage": "34942.53g / 11662.13m",
    "averageFilamentUsage": "40.54g / 13.53m",
    "highestFilamentUsage": "294.82g / 98.85m",
    "lowestFilamentUsage": "0.00g / 0.00m",
```
My code:
```
        "totalFilamentUsage": "46538.80g / 11662.13m",
        "averageFilamentUsage": "53.99g / 13.53m",
        "highestFilamentUsage": "417.80g / 98.85m",
        "lowestFilamentUsage": "0.00g / 0.00m",
```        
Any idea if Im being wrong? Could you check as well?

NotExpectedYet:
> have checked develop and the tool not showing up is a thing that happens there as well. This is code I kept intact, although I did have multiple questions why. Your code explicitly returns null for metrics being undefined. Maybe a follow-up PR is better.

Aye I was more meaning it shouldn't, but I think a follow up in the database is better like you suggested as it catches all those before they become an issue in the UI.

> **30 or 90 days ago** - @NotExpectedYet I found the reason why the graph was empty. Your variable thirtyDaysAgo was actually assigned a date of 90 days ago... so yeah my code was actually doing 30 days ago. Im at least I glad I found it. So now its all good with 90 days.
> 

Ahh! Forgot about that! Yeah 90 days is totally fine there, I'd originally set 30 because I thought it would kill it but nope. 

> **Loading** - About the loading... I made #549 and you made #538. So I think that is fixed in another PR.
> 

Agreed! 

> **Calculation** Final thing to note is that my code is outputting **higher usage numbers**. Let me show you:
> Previous code:
> 
> ```
>     "totalFilamentUsage": "34942.53g / 11662.13m",
>     "averageFilamentUsage": "40.54g / 13.53m",
>     "highestFilamentUsage": "294.82g / 98.85m",
>     "lowestFilamentUsage": "0.00g / 0.00m",
> ```
> 
> My code:
> 
> ```
>         "totalFilamentUsage": "46538.80g / 11662.13m",
>         "averageFilamentUsage": "53.99g / 13.53m",
>         "highestFilamentUsage": "417.80g / 98.85m",
>         "lowestFilamentUsage": "0.00g / 0.00m",
> ```
> 
> Any idea if Im being wrong? Could you check as well?

I'm just finishing off this settings PR and I will look. Off the top of my head it's either my original calculation was off, it's something with a state check, or it's something with the default fallback to PLA. The meters are fine, it's the gram calculation as your probably aware. So going to be one of those.

NotExpectedYet (approved):
Looks good, tested and all working this end. 

Congradulations, the first commit to 1.14-rc1

PR #410, opened 2021-04-03
Add Daemon Service prototype and innosetup work

davidzwa:
- [x] Innosetup + icon for setup file
- [x] Daemon service for systemd in .NET 5
- [ ] Tested daemon service on Ubuntu
- [ ] Icon file for daemon and octofarm executables (work in progress w.r.t. PKG exe file and dotnet publish adjustment)
- [ ] Daemon service installation code for windows service 

I was also thinking of adding special icons for dev/prerelease/alpha versions


davidzwa:
@NotExpectedYet abandoning this PR, not because it doesn't carry value but just because it is not that important for us right now. Its easy to reproduce. So, maybe one day.

NotExpectedYet:
> 
> 
> @NotExpectedYet abandoning this PR, not because it doesn't carry value but just because it is not that important for us right now. Its easy to reproduce. So, maybe one day.

I was thinking similar with the deb.

PR #419, opened 2021-04-07
Job clean refactor

davidzwa:
This is followup of #406 and will keep track of that first and when that other PR is merged, I will rebase this one on top.

- [x] Introduce luxon and date formatting
- [x] Cleanup JobClean (remove typeof undefined, awaits)
- [x] Test `generate(...)` and date formatting ✔️ ✔️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
     - Slightly reworked file cache, prepared for V2 and made it robust
     - Made API tests less prone to unnecessary failure
     - Reworked the Settings modal to be more resiliant to failure and cleaned up code. 
+    - Slightly reworked job cache, prepared for V2
 
 ### Removed
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "is-docker": "^2.1.1",
         "lodash": "^4.17.21",
         "lookpath": "^1.2.0",
+        "luxon": "^1.26.0",
         "mjpeg-decoder": "^1.0.3",
         "mongoose": "^5.12.9",
         "multer": "^1.4.2",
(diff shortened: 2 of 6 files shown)


davidzwa:
I introduce a certain date format here. I must stress that I see advantage in formatting dates **only client side**. That means just pass ISO date strings or Unix timestamps.

This goes further: I prefer not to render HTML server side, not to format numbers and not to provide units (although we could provide the unit of measurements separately). This way the client has the final say in formatting.
This is mainly important for V2.

NotExpectedYet:
> I introduce a certain date format here. I must stress that I see advantage in formatting dates **only client side** and me as well as Luxon advises against using custom formats. That means just pass ISO date strings or Unix timestamps and let the client figure it out.
> 
> This goes further: I prefer not to render HTML server side, not to format numbers and not to postfix (SI) units of measurement (although we could provide the unit of measurements separately). This way the client has the final say in formatting.
> This is mainly important for V2, but please take a moment to think about that in your refactoring work.
> If we can defer formatting as much as possible, I'd be a big fan.

Luxon is a sweet package good choice! 

Yes I agree there, we can certainly work to push this more client side. Especially for V2.

davidzwa:
@NotExpectedYet are the bugs you mentioned related to this PR? 
The 2nd: sure

> Some issues I've noticed, apart from these is looking good!
> 
> Expected filament costs.
> 
> 1. When no spool selected I get Total 0, and Tool 0: null. Total is correct but tool should not be null. We can't really default here regardless so 0 is fine.

I have no idea what to do haha. What should I look at in the UI or, even better, is this related to this PR?

> 2. Expected Units Total displays 0, even with units on Tool. This is actually a bug of mine so I'm happy to take a look if you'd prefer? I did this like. Been meaning to fix for a while.

Please do

> 3. Expected printer cost isn't fixed to 2 decimal places. Should be 0.04 on my test gcode and got 0.0429327937923

The third rounding bug is fixed for now server-side.

NotExpectedYet (approved):
I'll approve this now and then we can fix my bug in another PR.

PR #426, opened 2021-04-08
FileClean refactor

### LINKED ISSUE #425: FileClean cleanup
Title speaks for itself.

Reason: found some unhandy code (and Math.min/max bugs) and decided to test and study the code.

davidzwa:
- [x] Add test for FileClean
- [x] Upgrade FileClean and scan+refactor for weaknesses
- [x] Fix 2x max instead of min math bug
- [x] Changelog

Closes #425 
Draft until #419 is merged.

### DIFF
--- a/server_src/lib/dataFunctions/fileClean.js
+++ b/server_src/lib/dataFunctions/fileClean.js
@@ -2,7 +2,7 @@
 
 const _ = require("lodash");
 const Logger = require("../logger.js");
-const { getPrintCostNumeric } = require("../utils/print-cost.util");
+const {getPrintCostNumeric} = require("../utils/print-cost.util");
 
 const {
   getDefaultFileCleanStatistics
@@ -21,7 +21,7 @@ class FileClean {
     return fileStatistics;
   }
 
-  static async statistics(farmPrinters) {
+  static statistics(farmPrinters) {
     const storageFree = [];
     const storageTotal = [];
     const devices = [];
@@ -32,7 +32,7 @@ class FileClean {
 
     // Collect unique devices - Total for farm storage should not duplicate storage on instances running on same devices.
     farmPrinters.forEach((printer, index) => {
-      if (typeof printer.storage !== "undefined") {
+      if (!!printer.storage) {
         const device = {
           ip: printer.printerURL,
           index: printer.index,
@@ -40,8 +40,8 @@ class FileClean {
         };
         devices.push(device);
       }
-      if (typeof printer.fileList !== "undefined") {
-        printer.fileList.files.forEach((file) => {
+      if (!!printer.fileList) {
+        printer.fileList.files?.forEach((file) => {
           if (!isNaN(file.size)) {
             fileSizes.push(file.size);
           }
@@ -51,7 +51,7 @@ class FileClean {
 
           fileCount.push(file);
         });
-        printer.fileList.folders.forEach((file) => {
+        printer.fileList.folders?.forEach((file) => {
           folderCount.push(file);
         });
       }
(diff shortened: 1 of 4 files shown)


davidzwa:
This is dependent on history, but Ive decided not to stack it on top anymore... just to hard to keep up

NotExpectedYet:
Was thinking I could probably proceed with this at somepoint. Are we planning on keeping the builds in the repo? I haven't really thought much of it but presumed you'd want them building on git. 

I'll probably have to build the .deb in a similar way if so as this script is written for local building currently.

server_src/lib/dataFunctions/fileClean.js
```diff
@@ -75,100 +75,83 @@ class FileClean {
...
+        sortedFile.toolCosts = FileClean.getCost(
+          selectedFilament,
+          sortedFile.toolUnits,
+        );
+        sortedFileList.push(sortedFile);
       }
-      logger.info("File Information cleaned and ready for consumption...");
-      return true;
-    } catch (e) {
-      console.log("MASSIVE FILE ERROR!!!!", e);
```
davidzwa, line 161: Removed this, u ok with that?

NotExpectedYet: Yeah, is it caught somewhere if there are errors?

davidzwa: What kind of errors do we foresee?

NotExpectedYet: We could validate the file property from file list. path, fullPath, display, name, date, size, success,failed, last, time. 

There's not really as much as I'd thought going on here when I made the comment.

davidzwa: I started testing `statistics`, thats a bit more bug prone right now.

So in case an error occurs, I think saving a status variable and reason might help a lot in visualizing why certain cache's didnt generate. What do you think?

davidzwa:
@NotExpectedYet found a big bug (on my side) but I do really wonder how it passed your test.
The bug was on `FileClean.generate()` where `fileList` was iterated in a for-loop instead of `fileList?.files`. Its fixed now.

Secondly I found that `fileCount` needs to be passed to the cache, maybe better not to do that? (see my test file comment "// absurdism")

Thirdly, I see that state.js calls FileClean.generate immediately followed by FileClean.statistics a lot and passes farmPrinters each time.
image
Maybe we can reduce that, because this kinda stuff is heavy and recalculating it each time is very suboptimal as we have enough information on how to patch the state.

These are the calls, and for each I can picture clearly how to update the FileClean cache:
- setupWebSocket
- getFiles
- removeFile
- reSyncFile
- moveFile
- deleteFolder
- newFolder
- updateFile (2x, fileInformation.time or falsy)

server_src/lib/dataFunctions/fileClean.js
```diff
@@ -68,107 +68,113 @@ class FileClean {
...
+      return;
+    }
+
+    if (!!farmPrinter.systemChecks) {
+      farmPrinter.systemChecks.cleaning.file.status = "warning";
+    }
+    const printCost = farmPrinter.costSettings;
+    const sortedFileList = [];
+    if (!!fileList?.files) {
+      for (let file of fileList.files) {
```
davidzwa, line 121: This is the place of the bug I created and just fixed.

server_src/lib/dataFunctions/fileClean.js
```diff
@@ -68,107 +68,113 @@ class FileClean {
...
-      return true;
-    } catch (e) {
-      console.log("MASSIVE FILE ERROR!!!!", e);
     }
+
+    // TODO oof not the sortIndex again... this is SUPER BUG PRONE
+    // sortIndex should be a supporting math tool
+    cleanFileList[sortIndex] = {
+      fileList: sortedFileList,
+      filecount: fileList.fileCount || 0,
```
davidzwa, line 154: `filecount` is not camel case, this is more bug prone (for me) than you might think

server_src/lib/dataFunctions/fileClean.js
```diff
@@ -68,107 +68,113 @@ class FileClean {
...
+      folderList: fileList.folders || [],
+      folderCount: fileList.folderCount || 0,
```
davidzwa, line 156: I dont understand why we save folderCount and filecount if we can get the length from the array

server_src/lib/dataFunctions/fileClean.js
NotExpectedYet: Aye was a goof ages ago I never changed.

server_src/lib/dataFunctions/fileClean.js
NotExpectedYet: Was part of my "prep the client for everything" move with the cleaners.

davidzwa:
@NotExpectedYet lets get this merged. The try-catch which was removed I think should stay removed, reason: otherwise we'll never find bugs. All bugs in there are related to incorrect or missing properties, so I'd like things to break on dev and then fix it. 
We can work on try-catching on a higher level, so it doesnt break startup. Sound good?

NotExpectedYet:
> @NotExpectedYet lets get this merged. The try-catch which was removed I think should stay removed, reason: otherwise we'll never find bugs. All bugs in there are related to incorrect or missing properties, so I'd like things to break on dev and then fix it.
> We can work on try-catching on a higher level, so it doesnt break startup. Sound good?

Fine by me

PR #445, opened 2021-04-23
Feature/444+442 updater improvements

NotExpectedYet:
The updater will now check for missing dependencies and alert the user if there are new ones. 

The UI get's a response to ask if the user would like to continue with the installation of those dependencies, then the system will restart. 

Can't be tested with nodemon currently as the git pull makes the server restart and fails to send any response. 

If we're happy with the slow npm install command for now (shouldn't be too many situations with updated deps) then it's good for review.


NotExpectedYet:
Just to add tested `npm outdated --production` and only pulls productions dependencies :D

davidzwa:
> Just to add tested `npm outdated --production` and only pulls productions dependencies :D

Didnt look at the code, did you consider when there's no node_modules folder? Does that also work?

NotExpectedYet:
> > Just to add tested `npm outdated --production` and only pulls productions dependencies :D
> 
> Didnt look at the code, did you consider when there's no node_modules folder? Does that also work?

I did aye, it triggers the missing flag in the code regardless. The output is still the same so will always fire to the client. I could get fancy and tell them what is missing but didn't really see the relevance to them.

davidzwa (approved):
LGTM, untested: canary mode on

davidzwa (approved):
lo go to mo

Issue #439, opened 2021-04-19
Problem with printer data (temp, time)


3DRemake:
### To better help the developer understand your issue, please fill in the information below. Failing to do so will waste the developer and your own time in sorting the issue. Thanks!

**Client or Server side error** 
Client error
Uncaught (in promise) TypeError: Cannot set property 'innerHTML' of null
    at panelViewRunner.min.js:6504
    at u (serverAliveCheck.min.js:6554)
    at Generator._invoke (serverAliveCheck.min.js:6784)
    at Generator.next (serverAliveCheck.min.js:6609)
    at g (panelViewRunner.min.js:5864)
    at o (panelViewRunner.min.js:5866)
(anonymous) @ panelViewRunner.min.js:6504
u @ serverAliveCheck.min.js:6554
(anonymous) @ serverAliveCheck.min.js:6784
(anonymous) @ serverAliveCheck.min.js:6609
g @ panelViewRunner.min.js:5864
o @ panelViewRunner.min.js:5866
Promise.then (async)
g @ panelViewRunner.min.js:5864
o @ panelViewRunner.min.js:5866
(anonymous) @ panelViewRunner.min.js:5866
(anonymous) @ panelViewRunner.min.js:5866
(anonymous) @ panelViewRunner.min.js:5954

**Describe the bug**
If you don't connect the 1st printer of the list you can't see data for the second printer in panel tab or in camera tab (camera is ok, but data like time and temp are not ok)

**To Reproduce**
1. Have at least 2 printer registered, shutdown the octopi of the first printer in the list
2. Lanch a print on the second printer
3. Go to 'Panel' tab
4. See data abour temp and time are not refreshed

**Expected behavior**
Data should be refresh as they are if the first printer is connected

**Screenshots**
Screenshots of my second printer (First printer offline)
https://media.discordapp.net/attachments/709293261646200842/831923458534932530/unknown.png

Screenshots of my second printer (First printer online)
https://media.discordapp.net/attachments/709293261646200842/831926710332817438/unknown.png

Code snippet :

https://media.discordapp.net/attachments/709293261646200842/833694529273593876/unknown.png

**Versions (please complete the following information):**
 - NodeJS Version [12.22.1]
 - OctoFarm Version [1.1.12]
 - OctoPrint Version [1.5.3]
 - OctoPrint Plugins [N/A]

**Additional context**
N/A.

davidzwa:
[...]

PR #447, opened 2021-04-23
Attempted fix for #439

NotExpectedYet:
I couldn't reproduce this, but I've removed the pointless code and added checks in to make sure the tools exists before applying any value. 

I tried 
0: Fresh printer offline with no spool
1: Printer online and printering. 

0: Existing printer offline with no spool
1: Printer online and printering 

0: Existing printer offline with spool attached
1: printer online and printering

0: Fresh printer offline with spool attached
1: printer online and printering. 

Would like confirmation this is fixed from the user.


client_src/js/lib/modules/viewUpdater.js
```diff
@@ -1000,17 +1000,15 @@ async function updateState(printer, clientSettings, view) {
...
           tool.innerHTML = `${printer.selectedFilament[i].spools.material}`;
         }
       } else {
-        tool.innerHTML = "No Spool";
+        if (tool) {
+          tool.innerHTML = "No Spool";
+        }
       }
     }
-  } else {
```
NotExpectedYet, line 1012: This shouldn't even run so it's pointless. There's no tool boxes filled if it doesn't know the actual tool counts from the printer.

client_src/js/lib/modules/viewUpdater.js
```diff
@@ -1000,17 +1000,15 @@ async function updateState(printer, clientSettings, view) {
       const tool = document.getElementById(`${printer._id}-spool-${i}`);
       if (printer.selectedFilament[i] !== null) {
         const filamentManager = await checkFilamentManager();
-        if (filamentManager) {
```
NotExpectedYet, line 1003: Removed that pointless if function. Surely that was stoner laziness I'd be embarrassed if not XD

client_src/js/lib/modules/viewUpdater.js
```diff
@@ -1000,17 +1000,15 @@ async function updateState(printer, clientSettings, view) {
...
-          tool.innerHTML = `${printer.selectedFilament[i].spools.material}`;
-        } else {
+        if (tool) {
           tool.innerHTML = `${printer.selectedFilament[i].spools.material}`;
         }
       } else {
-        tool.innerHTML = "No Spool";
+        if (tool) {
```
davidzwa, line 1007: `else if` :P

NotExpectedYet: Nah can't be. The if is to protect the tool been updating the else is for selected filament been null

davidzwa: YOU SHALL NOT! be ... incorrect

davidzwa (approved):
No repro possible. Looks sorta ok to me

PR #450, opened 2021-04-23
Add in ability to restart OctoFarm from the database screen

NotExpectedYet:
Added in a restart endpoint that's unathenticated for the database issue screen. 

Button becomes activated when connection details is saved. 

Uses the already existing function in system

### DIFF
--- a/server_src/routes/databaseIssue.js
+++ b/server_src/routes/databaseIssue.js
@@ -8,13 +8,16 @@ const isDocker = require("is-docker");
 const envUtils = require("../utils/env.utils");
 const { AppConstants } = require("../app.constants");
 const { fetchMongoDBConnectionString } = require("../../app-env");
+const { SystemCommands } = require("../lib/serverCommands.js");
 
 const router = express.Router();
 const logger = new Logger("OctoFarm-Server");
 
 function validateMongoURL(mongoURL) {
   const mongoString = mongoURL.toLowerCase();
-  const hasMongoPrefix = mongoString.toLowerCase().includes("mongodb://") || mongoString.toLowerCase().includes("mongodb+srv://");
+  const hasMongoPrefix =
+    mongoString.toLowerCase().includes("mongodb://") ||
+    mongoString.toLowerCase().includes("mongodb+srv://");
   const hasOctoFarmTable = mongoString.includes("/octofarm");
 
   return {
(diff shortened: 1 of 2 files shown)


app-core.js
```diff
@@ -69,20 +69,28 @@ async function ensureSystemSettingsInitiated() {
...
 
   app.use("/", require("./server_src/routes/databaseIssue", { page: "route" }));
   app.use(
     "/serverChecks",
     require("./server_src/routes/serverChecks", { page: "route" })
   );
+  app.use(
+    "/settings",
+    require("./server_src/routes/unAthenticatedSettings", { page: "route" })
+  );
```
NotExpectedYet, line 93: New route containing the restart endpoint that's unathenticated

views/databaseIssue.ejs
```diff
@@ -100,6 +100,17 @@
...
+                            type="button"
+                            id="restart-server-button"
+                            class="btn btn-danger"
+                            onclick="postRestartServer()"
+                    <% if (isDocker) { %>
+                            hidden
+                            <% } %>
+                            disabled>
+                        Restart OctoFarm
+                    </button>
```
NotExpectedYet, line 113: New restart server button. Hidden for docker as this page won't be able to save anyway and the servers already restarted by the time they update their environment

views/databaseIssue.ejs
```diff
@@ -242,9 +254,38 @@
           messageElem.innerText = "Failed to save connection setting. Reason '" + b.reason + "'";
           messageElem.classList = ["text-danger"];
         }
-      });
+      })
+        .finally(() => {
+            getPageElements().restartOctoFarmButton.disabled = notYetValidatedConnection;
+        });
+  }
```
NotExpectedYet, line 261: Added a finally option after saving connection to enable the restart button

views/databaseIssue.ejs
```diff
@@ -242,9 +254,38 @@
...
+                  messageElem.innerText = "System restart command was successful,the server will restart in 5 seconds... Please refresh page or wait for automatic refresh in 10 seconds.";
+                  messageElem.classList = ["text-success"];
+                  setTimeout( () => location.reload(), 10 * 1000 )
+              }
+              else {
+                  const messageElem = getPageElements().connectionResponse;
+                  messageElem.innerText = "Failed to Restart OctoFarm. Reason:'" + b + "'";
+                  messageElem.classList = ["text-danger"];
+              }
+          });
```
NotExpectedYet, line 286: Pretty much followed your format here for the page. 

I don't return a message from the original function so just added it in as a string. 

I re-used your fetchOptions too as I dis-regard what's send in the body to the endpoint

server_src/routes/databaseIssue.js
```diff
@@ -156,4 +167,14 @@ router.post("/save-connection-env", async (req, res) => {
...
+router.post("/restart-octofarm", async (req, res) => {
+  let serviceRestarted = false;
+  try {
+    serviceRestarted = await SystemCommands.rebootOctoFarm();
+  } catch (e) {
+    logger.error(e);
+  }
+  res.send(serviceRestarted);
+});
+
```
NotExpectedYet, line 179: Did my own file at first then thought might as well just go in here rather than adding an additional route.

davidzwa (approved):
had to fix some

PR #451, opened 2021-04-24
bring back old issue report with updated note about new log dump

NotExpectedYet:
Ok so apparently github haven't released the yml issues to the wild yet :( I've applied for the repo to be included in the beta for them but this is just to bring the old file back in play until that happens.


davidzwa:
Ive been annoyed by this template a bit. If we return it, can it be shorter and more concise? My eyes no likey 👎🏼

davidzwa (approved):
LGTM, but do note I did overwrite the issue template.
Its now super compact and therefore more effective.

Also added dev task report as we use that a lot

PR #470, opened 2021-04-25
Bugfix/460 468 updater not errored out no docker update

NotExpectedYet:
Added in check for docker detection to not fire the update notification and also shows a note on the UI to refer to the website if confusion was effective. 

Switched out the .catch function for try{}catch(){} and the restart or update commands correctly bubble upwards and are now caught and returned to the client. 

Added Latest and Current version's with links to the UI


davidzwa (approved):
lgtm

PR #473, opened 2021-04-25
Fixed #472 - Latest release check, and fall back to current release

### LINKED ISSUE #472: Missed ?. from the new urls in system page doh!
- Description

Failing to produce the page due to missing ?.

- Repro

NotExpectedYet:
Just fix the page and add a fallback when object doesn't exist


davidzwa:
Applied `html_url` instead of `url` to refer the user to. Should be friendlier

davidzwa (approved):
- [x] Tested the containers with prerelease version
- [x] tested nodemon and corrected html_url

NotExpectedYet:
@davidzwa agree'd aha

PR #485, opened 2021-04-27
Feature/296 allow editing site title

NotExpectedYet:
Add's in an environment variable option to allow editing site title. 

Did not feel this needs adding automatically to the environment file. 

I think I hooked in right, just let me know if I missed anything


server_src/routes/index.js
```diff
@@ -43,6 +43,7 @@ async function welcome() {
       } else {
         res.render("welcome", {
           page: "Welcome",
+          octoFarmPageTitle: process.env.OCTOFARM_SITE_TITLE,
```
davidzwa, line 46: Try to use the process.env[AppConstants.OCTOFARM_SITE_TITLE_KEY] so we decouple the name from the variable intended

NotExpectedYet: Will PR now, aye was a bit but it works and it's in one place.

app-env.js
```diff
@@ -234,6 +234,18 @@ function getViewsPath() {
   return viewsPath;
 }
 
+function ensurePageTitle() {
+  if (!process.env[AppConstants.OCTOFARM_SITE_TITLE_KEY]) {
+    logger.info(
+        `~ ${AppConstants.OCTOFARM_SITE_TITLE_KEY} environment variable is not set. Assuming default: ${AppConstants.OCTOFARM_SITE_TITLE_KEY}=${AppConstants.defaultOctoFarmPageTitle}.`
+    );
+    printInstructionsURL();
```
davidzwa, line 242: I would not log this message and would avoid printing instructions. I think this needs to be a silent setting just like prerelease (which is gonna be removed, but still).

NotExpectedYet: I'm sure I removed that but will do another push for it aha

davidzwa: Ill do some pushups for too rapid rebasing!

NotExpectedYet: bahaha, I've lifted enough tables for the day does that count? 

Fixed anyway

davidzwa (changes requested):
image

davidzwa (approved):

NotExpectedYet:
> image

That's because off my merge conflict. The defaults always set. I'll fix.

davidzwa (approved):

PR #487, opened 2021-04-27
Node12 friendly abort 

davidzwa:
- [x] Add restart delay and backoff to `npm start` as well as entrypoint scripts.
- [x] Test with throw in app (quickest failure possible)
- [x] Test with test-node12.Dockerfile
- [x] Test with mono and alpy (windhose)
- [x] ~Fix .npmrc file instead of engine section~ both do not work

If necessary
- [x] Add friendly node12 abort msg at start

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,3 +1,7 @@
+if (parseInt(process.version.replace("v", "").split(".")[0]) < 14) {
+  throw "Please make sure Node 14 or higher is installed to run the latest OctoFarm.";
+}
+
 const { setupEnvConfig, fetchMongoDBConnectionString, fetchOctoFarmPort } = require("./app-env");
 const Logger = require("./server_src/lib/logger.js");
 const mongoose = require("mongoose");
--- a/app.js
+++ b/app.js
@@ -1,4 +1,10 @@
-if (parseInt(process.version.replace("v", "").split(".")[0]) < 14) {
+let majorVersion = null;
+try {
+  majorVersion = parseInt(process.version?.replace("v", "").split(".")[0]);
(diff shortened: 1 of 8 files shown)


app.js
```diff
@@ -1,3 +1,7 @@
+if (parseInt(process.version.replace("v", "").split(".")[0]) < 14) {
```
davidzwa, line 1: Try catch @NotExpectedYet maybe better in case process.version is not set?

docker/entrypoint.sh
```diff
@@ -14,11 +14,16 @@ then
     exit 1
 fi
 
+if [ -z "$OCTOFARM_PORT" ]
+then
+    echo "OCTOFARM_PORT=$OCTOFARM_PORT is not defined, the default of 4000 will be assumed. You can override this at any point."
```
davidzwa, line 19: Nice instruction for docker users. Personally I like it when containers log this kind instruction

davidzwa:
image

app.js
NotExpectedYet: Yeah I'd say so, there's always the chance.

package.json
```diff
@@ -30,7 +30,7 @@
     "pi",
     "node"
   ],
-  "author": "James Mackay (NotExpectedYet) & David Zwa",
+  "author": "James Mackay (NotExpectedYet) & David Zwart (davidzwa)",
```
NotExpectedYet, line 33: I prefered Zwa! XD

NotExpectedYet (approved):

PR #503, opened 2021-04-28
Bugfix/remove patreon

NotExpectedYet:
Just removing a recent patreon leaving

### DIFF
--- a/server_src/lib/serverCommands.js
+++ b/server_src/lib/serverCommands.js
@@ -8,6 +8,10 @@ const {
   doWeHaveMissingPackages,
   installMissingNpmDependencies,
 } = require("../utils/npm.utils.js");
+const {
+  checkIfFunctionExistsInPath,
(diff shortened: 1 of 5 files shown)


server_src/lib/serverCommands.js
```diff
@@ -19,23 +23,39 @@ class SystemCommands {
...
     if (isPm2()) {
-      try{
-        await exec("pm2 restart OctoFarm")
-        checkForNamedService = true;
-      }catch (e){
+      try {
+        let doesFunctionExist = await checkIfFunctionExistsInPath("pm2");
+
+        if (doesFunctionExist) {
+          setTimeout(async () => {
```
davidzwa, line 30: Shouldnt the timeout be awaited as well?

NotExpectedYet: Wrong commit anyway

PR #507, opened 2021-04-28
Hotfix/update improvements

NotExpectedYet:
Going to test it with a fake commit now

### DIFF
--- a/server_src/lib/serverCommands.js
+++ b/server_src/lib/serverCommands.js
@@ -58,7 +58,7 @@ class SystemCommands {
     const gitCurrentStatus = await returnCurrentGitStatus();
 
     // We can safely skip these checks if either of these are true as they should have been run by the time these get flagged.
-    if (force?.doWeInstallPackages && force?.forcePull) {
+    if (!(force?.doWeInstallPackages && force?.forcePull)) {
       // Check if branch is already up to date. Nothing to do, return response to user.
       const gitBranchUpToDate = isBranchUpToDate(gitCurrentStatus);
       if (gitBranchUpToDate) {
--- a/server_src/lib/serverCommands.js
+++ b/server_src/lib/serverCommands.js
@@ -11,6 +11,7 @@ const {
   isBranchInfront,
   doesBranchContainModifiedFiles,
   pullLatestRepository,
+  forcePullLatestRepository,
   checkIfWereInAGitRepo,
 } = require("../utils/git.utils.js");
 
@@ -41,19 +42,14 @@ class SystemCommands {
   }
 
   // This will need changing when .deb / installation script becomes a thing. It's built to deal with the current implementation.
-  static async checkIfOctoFarmNeedsUpdatingAndUpdate(force) {
-    let serverResponse = {
-      haveWeSuccessfullyUpdatedOctoFarm: false,
-      statusTypeForUser: "error",
-      message: "",
-    };
+  static async checkIfOctoFarmNeedsUpdatingAndUpdate(clientResponse, force) {
 
     // Check to see if current dir contains a git folder... hard fail otherwise.
     let isThisAGitRepo = await checkIfWereInAGitRepo();
     if (!isThisAGitRepo) {
-      serverResponse.message =
+      clientResponse.message =
         "Not a git repository, user intervention required! You will have to re-download OctoFarm and re-unpack it over this directory. Make sure to backup your images folder!";
-      return serverResponse;
+      return clientResponse;
     }
 
     const gitCurrentStatus = await returnCurrentGitStatus();
@@ -63,11 +59,11 @@ class SystemCommands {
       // Check if branch is already up to date. Nothing to do, return response to user.
       const gitBranchUpToDate = isBranchUpToDate(gitCurrentStatus);
       if (gitBranchUpToDate) {
-        serverResponse.haveWeSuccessfullyUpdatedOctoFarm = false;
-        serverResponse.message =
+        clientResponse.haveWeSuccessfullyUpdatedOctoFarm = false;
+        clientResponse.message =
           "OctoFarm is already up to date! Your good to go!";
-        serverResponse.statusTypeForUser = "success";
-        return serverResponse;
+        clientResponse.statusTypeForUser = "success";
+        return clientResponse;
       }
 
       //The below checks should only run if the branch is not up to date... pointless otherwise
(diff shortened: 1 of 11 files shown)


server_src/lib/serverCommands.js
```diff
@@ -19,138 +26,110 @@ class SystemCommands {
...
-        try{
-          await exec("git pull")
-        }catch(e){
-          serverResponse.message = `Could not pull the latest files, user intervention required | Error: ${stderr}`;
-          throw serverResponse;
-        }
+    const gitCurrentStatus = await returnCurrentGitStatus();
+
+    // We can safely skip these checks if either of these are true as they should have been run by the time these get flagged.
+    if (!(force?.doWeInstallPackages && force?.forcePull)) {
```
davidzwa, line 61: or logic is a bit clearer

server_src/utils/git.utils.js
```diff
@@ -0,0 +1,39 @@
...
+  return status.behind === 0;
+}
+
+function isBranchInfront(status) {
+  return status.ahead !== 0;
+}
+
+function doesBranchContainModifiedFiles(status) {
+  if (status.modified.length !== 0) {
+    return status.modified;
```
davidzwa, line 18: Return type inconsistent. Should be boolean, so convert with !!

server_src/utils/git.utils.js
```diff
@@ -0,0 +1,39 @@
...
+  } else {
+    return false;
+  }
+}
+
+async function pullLatestRepository(force) {
+  if (force.forcePull) {
+    return git.pull();
+  } else {
+    await git.reset("hard");
```
davidzwa, line 28: Why not provide forcePull directly?
Also, why does forcePull false trigger a reset? Not correct or unclear to me

server_src/lib/serverCommands.js
```diff
@@ -8,6 +8,13 @@ const {
...
 } = require("../utils/npm.utils.js");
+const {
+  returnCurrentGitStatus,
+  isBranchUpToDate,
+  isBranchInfront,
+  doesBranchContainModifiedFiles,
+  pullLatestRepository,
+} = require("../utils/git.utils.js");
 
 function isGitSync(dir) {
```
davidzwa, line 19: Should be removed with simple-git.

server_src/utils/git.utils.js
NotExpectedYet: No because I want the modified file list

server_src/lib/serverCommands.js
NotExpectedYet: Done

server_src/utils/git.utils.js
NotExpectedYet: That was a mistake, I've corrected the boolean check. 

What do you mean provide it directly? put it in it's own function? Don't mind either way.

davidzwa: `function someFunction(forcePull)` instead of `function someFunction(force)`

NotExpectedYet: Sorted

test/application/server-commands.test.js
```diff
@@ -0,0 +1,95 @@
...
+                },
+                checkIsRepo: () => true
+            }
+        }
+    });
+
+    jest.mock("../../server_src/utils/npm.utils", () => {
+        return () => {
+            return {
+                doWeHaveMissingPackages: () => Promise.resolve(true),
```
davidzwa, line 48: Make this dynamic with a variable. Same idea as currentScenario above

test/application/server-commands.test.js
```diff
@@ -0,0 +1,95 @@
...
+const git = simpleGit();
+
+const scenarioUpToDate = {
+    modified: [],
+    ahead: 0,
+    behind: 0,
+}
+
+describe('ServerCommands', () => {
+    const scenarioModifiedOutput = {
```
davidzwa, line 11: Add more dynamic scenario's here and switch simple-git's output before testing each scenario.

test/application/server-commands.test.js
```diff
@@ -0,0 +1,95 @@
...
+    };
+
+    jest.mock("child_process", () => {
+        return {
+            exec: () => Promise.resolve()
+        }
+    });
+
+    let wasReset = null;
+    let currentScenario = scenarioUpToDate;
```
davidzwa, line 28: currentScenario is non-constant, so we can override it at any time.

test/application/server-commands.test.js
```diff
@@ -0,0 +1,95 @@
...
+    jest.mock("simple-git", () => {
+        return () => {
+            return {
+                status: () => Promise.resolve(
+                    currentScenario
```
davidzwa, line 33: The last set scenario is picked up and provided as internal output of simple-git

test/application/server-commands.test.js
```diff
@@ -0,0 +1,95 @@
...
+                ),
+                pull: () => Promise.resolve(),
+                reset: (input) => {
+                    wasReset = true;
```
davidzwa, line 37: Nice marker variable (like a spy) to indicate whether it was called or not.

test/application/server-commands.test.js
```diff
@@ -0,0 +1,95 @@
...
+        const serverResponse = await SystemCommands.checkIfOctoFarmNeedsUpdatingAndUpdate( {},true);
+        expect(serverResponse.message).toBe("OctoFarm is already up to date! Your good to go!");
+        expect(serverResponse.haveWeSuccessfullyUpdatedOctoFarm).toBe(false);
+        expect(serverResponse.statusTypeForUser).toBe('success');
+    });
+
+    it("should be able to detect and fix uninstalled packages", async () => {
+    });
+
+    it("should be able to detect and fix uninstalled packages", async () => {
```
davidzwa, line 66: - scenario for when we dont have missing packages, but unstaged changes **without** force
- scenario for when we dont have missing packages, but unstaged changes **with** force
- scenario for when git pull throws that git is not installed (just an idea?)
- scenario for when git does not connect (air-gapped?)
- scenario for when we dont have either (good to go)

More ideas?

test/application/server-commands.test.js
```diff
@@ -0,0 +1,95 @@
...
+    const {SystemCommands} = require("../../server_src/lib/serverCommands");
+
+    it("should be able to see non-pushed commits", async () => {
+        const serverResponse = await SystemCommands.checkIfOctoFarmNeedsUpdatingAndUpdate( {},true);
+        expect(serverResponse.message).toBe("OctoFarm is already up to date! Your good to go!");
+        expect(serverResponse.haveWeSuccessfullyUpdatedOctoFarm).toBe(false);
+        expect(serverResponse.statusTypeForUser).toBe('success');
+    });
+
+    it("should be able to detect and fix uninstalled packages", async () => {
```
davidzwa, line 63: scenario when we are missing packages. Make doWeHaveMissingPackages() above dynamic using a state variable right above. That way you can set the value before the mocked function is called.

test/application/server-commands.test.js
```diff
@@ -0,0 +1,95 @@
...
+    });
+
+    it("should be able to detect and fix uninstalled packages", async () => {
+    });
+
+    it("should be able to see that we have a git repo", async () => {
+        const isRepo = await git.checkIsRepo();
+        expect(isRepo).toBe(true);
```
davidzwa, line 71: We have a git repo for this test scenario. This should fail in PKG mode, ZIP archive mode and in docker mode.
Nice to test those differently.

test/application/server-commands.test.js
```diff
@@ -0,0 +1,95 @@
...
+    });
+
+    it("should not reboot octofarm in unknown mode", async () => {
+        // Output indicates that we are neither in pm2 or nodemon mode
+        const output = await SystemCommands.rebootOctoFarm();
+        expect(output).toBe(false);
```
davidzwa, line 77: false indicates that the restart is not possible in this setup.

test/domain-shared/env.test.js
```diff
@@ -0,0 +1,5 @@
+describe("NODE_ENV", () => {
+    it("environment should be test", () => {
+        expect(process.env.NODE_ENV).toBe("test");
```
davidzwa, line 3: Just nice to know that we can always depend on NODE_ENV to be development, production or test.
These can be used to differentiate logging filters, or level of detail for exceptions caught in an exception handler.

test/domain-shared/git-utils.test.js
```diff
@@ -0,0 +1,65 @@
...
+            "package.json",
+            "server_src/lib/serverCommands.js",
+        ],
+        ahead: 0,
+        behind: 0,
+    };
+
+    let wasReset = null;
+
+    jest.mock("simple-git", () => {
```
davidzwa, line 14: Similar to system-command tests, but I'd suggest to keep a copy locally as each test is different.

test/domain-shared/git-utils.test.js
```diff
@@ -0,0 +1,65 @@
...
+        };
+    });
+
+    const {
+        returnCurrentGitStatus,
+        isBranchUpToDate,
+        isBranchInfront,
+        getListOfModifiedFiles,
+        pullLatestRepository,
+    } = require("../../server_src/utils/git.utils");
```
davidzwa, line 33: `require` after `jest.mock`. 
That's how jest is able to inject the mocked function, otherwise the mock would not be inserted and the real call would still happen. This is also a nice way to switch back and forth to a real implementation sometimes.

test/domain-shared/git-utils.test.js
```diff
@@ -0,0 +1,65 @@
...
+        expect(gitResponse).toBeTruthy();
+        expect(gitResponse.modified).toHaveLength(3);
+    });
+
+    it("should be able to reset changes when force is true", async () => {
+        await pullLatestRepository(false);
+        expect(wasReset).toBe(null);
+        wasReset = null;
+        await pullLatestRepository(true);
+        expect(wasReset).toBe(true);
```
davidzwa, line 63: Force: true => git reset is done. Thats why wasReset is expected to be true, we set it explicitly in our mock above.

davidzwa (approved):

test/domain-shared/git-utils.test.js
NotExpectedYet: Aye noticed like just went to cook food aha. Cheers for sorting it.

PR #516, opened 2021-04-30
Bugfix/513 no file information

NotExpectedYet:
This is a pickler... not sure what's happened. 

I've cleaned up some initial file object checks I'd screwed up. The thing is they should be there, and not null. The user would see no printing files so I feel he's been printing from OctoPrint and left out some information or didn't see it relevant. 

I have not been able to reproduce the issue here again! 

I have tried deleting all files so fileList is empty, and setting current.data.job to null so there's no information to go from. Apart from that breaking history showing anything it didn't produce the cleaner error. 

I didn't log the error for the try / catch on printerClean in my infinite wisdom so I've nothing really to go off around that. The only thing I had to work from we're the Errors in his logs, which I have now fixed in this PR with null checks on those values. 

A restart not clearing this error is most disconcerting, it would lead me to believe it's something with the initial scan which the only issue I saw was a file scan not completing because it took too long. These are multiple instances on WiFi I gather so who knows if that's actually going to respond in a quick enough time. 

Need user to test further to confirm.


davidzwa:
Contains duplicate code compared to #406 also that branch contains tests, so if this is the wish I think that deserves our attention more.

NotExpectedYet:
Well I'll change this PR to just include the client changes and bolster the printer view to show the printer if error'd out at least so a user can inspect more in the logs. 

Your PR I'll take a look at and see if we can get it in this hotfix for 1.1.13?

PR #517, opened 2021-04-30
Duplicate username regression

davidzwa:
- [x] Adds uniqueness to username schema
- [x] Tests it to throw on duplicate usernames
- [x] Profit

Please review by stating what we need to test more. 
- This schema change can cause duplicates in database to crash at the first `find()` call
- deDup is apparently one possibility to apprehend that

### DIFF
--- a/server_src/models/User.js
+++ b/server_src/models/User.js
@@ -7,7 +7,8 @@ const UserSchema = new mongoose.Schema({
     },
     username: {
         type: String,
-        required: true
+        required: true,
+        unique: true
     },
     password: {
         type: String,
--- a/test/domain/user-schema.test.js
+++ b/test/domain/user-schema.test.js
@@ -1,0 +1,62 @@
+const dbHandler = require("../db-handler");
+const User = require("../../server_src/models/User");
+const bcrypt = require("bcryptjs");
+
+/**
+ * Connect to a new in-memory database before running any tests.
+ */
+beforeAll(async () => {
+    await dbHandler.connect();
+});
+
+/**
+ * Clear all test data after every test.
+ */
+afterEach(async () => {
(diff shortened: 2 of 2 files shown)


NotExpectedYet:
Do we have a check and a response to the user if a duplicate username is attempted or does it just hard fail? 

I gather it would just create the first in this situation and continue through as normal without the duplicate.

davidzwa:
As the error is caught and logged with console.log(err) nothing will be presented. That's the what the `/users/register` route looks like right now.

NotExpectedYet:
Could do with updating then really as well if that's the case. There's a flash we can use that checks the passwords and such shouldn't be too hard to check for duplicated user. 

I gather it's my addition of the length check for creating the administrator group that caused this anyway.

davidzwa:
We do not know yet, so all we can do is make sure that db is consistent by constraint and regression testing.

NotExpectedYet:
I'll need to test the case of what happens with two users in the dB as well just incase that eventuality is out there.

davidzwa:
Is something we can test by removing and readding the index.
Tbh its pretty clear to me that we must think of deduplication

NotExpectedYet (approved):
Fine with this. Just tried it by duplicating my current Administrator user and it's no errors or issues are thrown. I can also log into my first user.

davidzwa:
Hmm that's weird because your database should be throwing an error when you create duplicate usernames... That's why I started about `deDup`

NotExpectedYet:
@davidzwa Not sure, it's still there too and going fine.

NotExpectedYet:
@NotExpectedYet I didn't create btw, I duplicated it on compass then ran your PR.

Issue #532, opened 2021-05-02
OP printer settings dialog does not query fresh settings


davidzwa:
- Description
Upon editing settings we gotta have the latest settings set from OP. Fix that with an OP API call. Also add a loader to printer settings dialog.

- Repro
Open printer settings dialog.

davidzwa:
It now queries the settings, but its not passed on to the UI.

```
09/05/2021, 10:09:18 | INFO | OctoFarm-State | Attempting to connect to API: api/system/commands | http://192.168.1.177 | timeout: 10000
09/05/2021, 10:09:18 | INFO | OctoFarm-State | Successfully grabbed System Information for...: http://192.168.1.177
09/05/2021, 10:09:35 | INFO | OctoFarm-State | Attempting to connect to API: api/settings | http://192.168.1.177 | timeout: 10000
09/05/2021, 10:09:35 | INFO | OctoFarm-State | Detected Pi Support!
09/05/2021, 10:09:35 | INFO | OctoFarm-State | Attempting to connect to API: api/plugin/pi_support | http://192.168.1.177 | timeout: 10000
09/05/2021, 10:09:35 | INFO | OctoFarm-State | Got from endpoint: data:{"model":"Raspberry Pi 4 Model B Rev 1.2","model_unrecommended":false,"throttle_state":{"current_issue":false,"current_overheat":false,"current_undervoltage":false,"past_issue":false,"past_overheat":false,"past_undervoltage":false,"raw_value
":-1}} |
```

NotExpectedYet:
> It now queries the settings, but its not passed on to the UI.
> 
> ```
> 09/05/2021, 10:09:18 | INFO | OctoFarm-State | Attempting to connect to API: api/system/commands | http://192.168.1.177 | timeout: 10000
> 09/05/2021, 10:09:18 | INFO | OctoFarm-State | Successfully grabbed System Information for...: http://192.168.1.177
> 09/05/2021, 10:09:35 | INFO | OctoFarm-State | Attempting to connect to API: api/settings | http://192.168.1.177 | timeout: 10000
> 09/05/2021, 10:09:35 | INFO | OctoFarm-State | Detected Pi Support!
> 09/05/2021, 10:09:35 | INFO | OctoFarm-State | Attempting to connect to API: api/plugin/pi_support | http://192.168.1.177 | timeout: 10000
[...]

Issue #531, opened 2021-05-02
OctoPrint port setting - actual saved value could be checked against available ports


davidzwa:
- Description
The setting for `port` is not checked, which could give OF information about what is saved in OP and what can be done to fix it for a user.

- Repro
Go to a printer settings dialog and check the port is empty because the port is not listed.

PR #545, opened 2021-05-03
Feature/printer settings dialog improvements

NotExpectedYet:
Printer Settings Class refactor. 

Cleaned up a lot of code so far, still some things left to do. 

Using this as the base to start the work on #532 #531

 - [x] Replace functions that load the settings modal. 
 - [x] Refactor the code to load everything under it's own static function
 - [x] Add in try/catches and respond the error to the client if a problem arises
 - [x] Create a new octofarm client file and improve the functions with error responses in a single place
 - [x] Add in new api endpoint to trigger and update for a printers settings from octoprint
 - [x] Make settings panel trigger the update from OctoFarm for octoprint's latest settings. 
 - [x] Replace global try/catch with multiple so 1 panel failing to load won't break the rest
 - [x] Put back in offline checks into panels
 - [x] Put back in offline message
 - [x] Add in check for port not available in drop down
 - [x] Add back in listener for save function
 - [x] Refactor the save function
 - [x] Fix save button event listener not applying
 - [x] Add kill button event listener
 - [x] Add test for endpoint
 - [x] Add changelog
 - [x] Add in port refresh
 - [x] Combine response into 1 alert.
 - [x] Add page refresh after save
 - [x] Clean up the extra calls on settings saving. Client refresh will make these redundant.


server_src/routes/printers.js
```diff
@@ -102,7 +102,7 @@ router.post("/updateSettings", ensureAuthenticated, async (req, res) => {
   const settings = req.body;
   logger.info("Update printers request: ", settings);
   const updateSettings = await Runner.updateSettings(settings);
-
+  res.sendStatus(200);
```
davidzwa, line 105: Why should we set a status if default is 200? Im a lazy .NET dev, all my status codes are automagically set XD

views/printerManagement.ejs
```diff
@@ -564,27 +564,27 @@
                     <div class="row">
                         <div class="col-12 col-sm-12 col-md-12 col-lg-2 col-xl-2">
                             <div class="list-group" id="list-tab" role="tablist">
-                                <a class="list-group-item list-group-item-action active" id="printer-connection-btn"
+                                <a class="list-group-item list-group-item-action active notyet" id="printer-connection-btn"
```
davidzwa, line 567: What does this class name do?

server_src/routes/printers.js
NotExpectedYet: I think this can be removed, I was chasing the settings issue and thought it was here. Will test and remove. Thought it was sending a 204 for some reason.

views/printerManagement.ejs
NotExpectedYet: disables an a tag

server_src/routes/printers.js
davidzwa: Nothing wrong with a 204: just means the body is empty.

NotExpectedYet: I want a body, I specifically look for 204's in the new OctoFarm client.

davidzwa (changes requested):
The PR does not fix #532, so my proposal is to remove that from the linked issues. The reason for this is because I have a more effective solution for it in mind for Printer Clean (for which it would be great if #406 is merged first, that would help me not having to pile a PR on top of another PR like file and job clean).

If you fix the event listener, remove the issue from linked issues I think this PR is good to go!

Dont forget:
- changelog
- Jest tests

NotExpectedYet:
I'll undo the mystery machine 🤣

NotExpectedYet:
Done the redirect mock, but not a clue with mocking passport.js at moment will look later when I have more time.

client_src/js/lib/functions/ui.js
```diff
@@ -89,4 +89,17 @@ export default class UI {
...
+        '<i class="fas fa-spinner fa-spin"></i>',
+        ""
+      );
+    }
+  }
+  static addLoaderToElementsInnerHTML(element) {
+    if (!element.innerHTML.includes("spinner")) {
+      element.innerHTML += ' <i class="fas fa-spinner fa-spin"></i>';
+    }
+  }
```
NotExpectedYet, line 104: Just tow new UI functions that I use quite a lot. Rest of client can start utilising these more but will migrate in another PR.

client_src/js/lib/octofarm_client.js
```diff
@@ -0,0 +1,42 @@
...
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify(data),
+    });
+    if (post.status === 400) {
+      throw new Error(
+        `Malformed request! Status: ${post.status} - ${post.statusText}`
+      );
+    } else if (post.status === 503) {
```
NotExpectedYet, line 34: Created a new OctoFarm client that does the status checks and then passes back json if available. Simplifying the code required to utilising it and stopping all the repeated status checks elseware. Plan to migrate rest of code in another PR but left the old one in for now.

client_src/js/printerManagementRunner.js
```diff
@@ -3,7 +3,7 @@ import OctoPrintClient from "./lib/octoprint.js";
 import OctoFarmClient from "./lib/octofarm.js";
 import UI from "./lib/functions/ui.js";
 import PrinterManager from "./lib/modules/printerManager.js";
-import PrinterSettings from "./lib/modules/printerSettings.js";
+import { updatePrinterSettingsModal } from "./lib/modules/printerSettings.js";
```
NotExpectedYet, line 6: Rather than importing the full class I decided to just create a function that applied the state and utilised the statics in the file.

client_src/js/printerManagementRunner.js
```diff
@@ -94,11 +91,23 @@ if (window.Worker) {
...
   }
 } else {
   // Sorry! No Web Worker support..
-  console.log("Web workers not available... sorry!");
+  UI.createAlert(
+    "error",
+    "Sorry web workers are not supported in your browser, please check the supported browser list.",
+    0
+  );
+  console.error(`Web workers not available... sorry! <br> ${e}`);
```
NotExpectedYet, line 110: Updated these with a UI alert rather than hiding in console.

client_src/js/printerManagementRunner.js
```diff
@@ -3321,7 +3329,7 @@ class dashUpdate {
             document
               .getElementById(`printerStatistics-${printer._id}`)
               .addEventListener("click", async (e) => {
-                PrinterLogs.loadStatistics(printer._id);
+                await PrinterLogs.loadStatistics(printer._id);
```
NotExpectedYet, line 3332: Missing await

server_src/lib/dataFunctions/printerClean.js
```diff
@@ -61,8 +61,15 @@ class PrinterClean {
...
-    return printersInformation;
+  static returnPrintersInformation(id) {
+    if (!id) {
+      return printersInformation;
+    } else {
+      const index = _.findIndex(printersInformation, function (o) {
+        return o._id == id;
+      });
+      return printersInformation[index];
+    }
```
NotExpectedYet, line 72: Updated this function to work with and without a specific printer. No id will return the entire list, and id provided will pull out the requested printer. This will simplify the code elseware a lot with the check done here. Follow up PR to update to utilise this more but will await the cache finishing for printers.

server_src/routes/printers.js
```diff
@@ -9,10 +9,14 @@ const { Runner } = runner;
...
 
 const logger = new Logger("OctoFarm-API");
+
 const printerClean = require("../lib/dataFunctions/printerClean.js");
 
 const { PrinterClean } = printerClean;
 
+// Doesn't returns undefined, note to use is incorrect...
+// const { returnPrintersInformation } = require("../cache/printer.cache.js");
+
```
NotExpectedYet, line 19: Just a comment when I was testing, can be removed when cache is implemented.

server_src/routes/printers.js
```diff
@@ -98,14 +102,12 @@ router.post("/feedChange", ensureAuthenticated, async (req, res) => {
...
   // Check required fields
-
   const settings = req.body;
   logger.info("Update printers request: ", settings);
   const updateSettings = await Runner.updateSettings(settings);
-
   res.send({ status: updateSettings.status, printer: updateSettings.printer });
 });
-router.get("/killPowerSettings/:id", ensureAuthenticated, async (req, res) => {
+router.post("/killPowerSettings/:id", ensureAuthenticated, async (req, res) => {
```
NotExpectedYet, line 110: Updated to a post command as backend is changed

server_src/routes/printers.js
```diff
@@ -125,25 +127,26 @@ router.get("/groups", ensureAuthenticated, async (req, res) => {
 });
 router.post("/printerInfo", ensureAuthenticated, async (req, res) => {
   const id = req.body.i;
+  const returnPrinter = await PrinterClean.returnPrintersInformation(id);
+  res.send(returnPrinter);
+});
```
NotExpectedYet, line 132: Simplified the code with the new allowance of an ID check within return printers information

server_src/routes/printers.js
```diff
@@ -125,25 +127,26 @@ router.get("/groups", ensureAuthenticated, async (req, res) => {
...
+  try {
+    await Runner.getLatestOctoPrintSettingsValues(id);
+    let printerInformation = PrinterClean.returnPrintersInformation(id);
+    res.send(printerInformation);
+  } catch (e) {
+    logger.error(`The server couldn't update your printer settings! ${e}`);
+    res.statusMessage = `The server couldn't update your printer settings! ${e}`;
+    res.sendStatus(500);
   }
 });
```
NotExpectedYet, line 151: New endpoint to update printer settings, must provide an ID for it. Sends information to client when no id provided with status message. I would like to utilise these status messages a lot more as it will help people using the client in the future to actually get an error response back rather than just a status code.

server_src/runners/state.js
```diff
@@ -3000,6 +2998,23 @@ class Runner {
...
+    // Update the printers cached settings from OctoPrint
+    await Runner.getSettings(id);
+    // Update the printers cached system settings from OctoPrint
+    await Runner.getSystem(id);
+    // Re-generate the printer clean information - This is just cautionary, my tests showed it wasn't needed.
+    await PrinterClean.generate(
+      farmPrinters[index],
+      systemSettings.filamentManager
+    );
+  }
```
NotExpectedYet, line 3017: New function for state. Just runs through the required calls to OctoPrint that are needed to update the printer settings on OctoFarms side. The extra generate wasn't needed but I'd rather keep it just encase there are timing issues server side.

NotExpectedYet: Can be cleaned up more once the refactor of state JS has happened.

test/api/printers-check.test.js
```diff
@@ -0,0 +1,50 @@
...
+  it("should return 302 redirect when no user is logged in", async function (done) {
+    process.env.npm_package_version = require("../../package.json").version;
+    process.env.testlatest_package_version = require("../../package.json").version;
+    app = await getOrCreateApp();
+    const res = await request(app)
+      .post("/printers/updatePrinterSettings")
+      .send();
+    expect(res.statusCode).toEqual(302);
+    done();
+  }, 10000);
```
NotExpectedYet, line 49: Initial test for redirect, but will need to update once I figure out how to do passport auth

views/printerManagement.ejs
```diff
@@ -564,32 +564,31 @@
...
                                    data-toggle="list" href="#printer-cost" role="tab" aria-controls="messages"><i
-                                            class="fas fa-dollar-sign"></i> Cost</a>
-                                <a class="list-group-item list-group-item-action" id="printer-profile-btn"
+                                            class="fas fa-dollar-sign"></i> Cost <i class="fas fa-spinner fa-spin"></i></a>
+                                <a class="list-group-item list-group-item-action notyet" id="printer-profile-btn"
                                    data-toggle="list" href="#printer-profile" role="tab" aria-controls="messages"><i
-                                            class="fas fa-tags"></i> Profile</a>
-                                <a class="list-group-item list-group-item-action" id="printer-power-btn"
+                                            class="fas fa-tags"></i> Profile <i class="fas fa-spinner fa-spin"></i></a>
+                                <a class="list-group-item list-group-item-action notyet" id="printer-power-btn"
```
NotExpectedYet, line 576: notyet is a specific class I created to disable the action on a links. It's removed from here once the section has loaded. It's handy if you want to disable an "a" link as they don't work with disabled as buttons do.

client_src/js/lib/modules/printerSettings.js
```diff
@@ -1,1325 +1,1403 @@
...
-let currentIndex = 0;
-
-// let controlDropDown = false;
-
-let currentPrinter = null;
+let currentPrinterIndex;
+let printerOnline;
+let currentPrinter;
+let hasErrorNotificationBeenTriggered = false;
+let pageElements;
```
NotExpectedYet, line 11: global elements to store page information. 
CurrentPrinterIndex = used when the sse event fires again so we don't have to keep checking it. Number
printerOnline = boolean for status check on printers state. This is relevant to the host status. Not the printer status
currentPrinter = Object containing the current printer
hasErrorNotificationBeenTriggered = boolean that get's updated so the UI alert doesn't keep firing for sse event
pageElements = holder for the page elements

client_src/js/lib/modules/printerSettings.js
```diff
@@ -1,1325 +1,1403 @@
...
-  // if (document.getElementById("printerSelection")) {
-  //   document.getElementById("printerSelection").remove();
-  //   controlDropDown = false;
-  // }
-});
 $("#connectionModal").on("hidden.bs.modal", function (e) {
   if (document.getElementById("connectionAction")) {
     document.getElementById("connectionAction").remove();
   }
 });
```
NotExpectedYet, line 18: Listens for the modal closing and removes the connectionAction element so it can be replaced.

client_src/js/lib/modules/printerSettings.js
```diff
@@ -1,1325 +1,1403 @@
...
 
-export default class PrinterSettings {
-  static async init(index, printers, printerControlList) {
-    try {
-      if (index !== "") {
-        UI.clearSelect("ps");
-        const printerProfileBtn = document.getElementById(
-          "printer-profile-btn"
+async function updatePrinterSettingsModal(printersInformation, printerID) {
```
NotExpectedYet, line 20: This function is fired initially from the API call. It receives the printer ID and updates the rest of the page with that specific printer. 
It's then fired again from the SSE event, without a printer ID this time. The variables above will be filled by this time with the specific printers ID so we can continue updating the live elements. It's mainly the printer status stuff at the bottom of the page.

client_src/js/lib/modules/printerSettings.js
```diff
@@ -1,1325 +1,1403 @@
...
+              <td>${s.message}</td>
+            </tr>
+
+        `
+      );
+      document.getElementById(`activePrinter-${s._id}`).checked = s.active;
+      const triggerSelect = document.getElementById(`triggerPrinter-${s._id}`);
+      triggerSelect.innerHTML = await Script.alertsDrop();
+      triggerSelect.value = s.trigger;
+    });
```
NotExpectedYet, line 853: Just trying out .map here rather than for each. Depends what your doing really but this can be faster and more resilient.

client_src/js/lib/modules/printerSettings.js
```diff
@@ -1,1325 +1,1403 @@
...
-    elements.mainPage.status.className = `btn btn-${printer.printerState.colour.name} mb-1 btn-block`;
-    elements.mainPage.host.innerHTML = `<b>Host Status</b><br>${printer.hostState.state}`;
-    elements.mainPage.host.className = `btn btn-${printer.hostState.colour.name} mb-1 btn-block`;
-    elements.mainPage.socket.innerHTML = `<b>WebSocket Status</b><br>${printer.webSocketState.desc}`;
-    elements.mainPage.socket.className = `btn btn-${printer.webSocketState.colour} mb-1 btn-block`;
+  static disablePanel(element) {
+    if (element.classList.contains("notyet")) {
+      element.classList.remove("notyet");
+    }
+  }
```
NotExpectedYet, line 1275: I may move these to UI once more places want to utilise them. Handy functions.

client_src/js/lib/modules/printerSettings.js
```diff
@@ -1,1325 +1,1403 @@
...
-        printerGcodeBtn.classList.add("notyet");
+      if (!pageElements.menu.printerGcodeBtn.classList.contains("notyet")) {
+        pageElements.menu.printerGcodeBtn.classList.add("notyet");
       }
-      if (!printerOtherSettings.classList.contains("notyet")) {
-        printerOtherSettings.classList.add("notyet");
+      if (
+        !pageElements.menu.printerOtherSettings.classList.contains("notyet")
+      ) {
+        pageElements.menu.printerOtherSettings.classList.add("notyet");
```
NotExpectedYet, line 1396: If the printer goes offline whilst your on this page, this will make the UI react.

client_src/js/lib/functions/ui.js
NotExpectedYet: The remove from loader was also causing the listener to disappear which is something I'll have to keep in mind. Initially there was no check if "spinner" was included in the innerHTML so it was replacing nothing basically. That was causing the listener to disappear which I had no idea was an effect of that.

NotExpectedYet (commented):
Annotated.

davidzwa:
> Done the redirect mock, but not a clue with mocking passport.js at moment will look later when I have more time.

I wouldnt mock passport. I'd mock the `ensureCurrentUserAndGroup` and `ensureAuthenticated`

NotExpectedYet:
I've updated the 400 test and also added the 500 test. 

Is it worth doing the test for Runner at the moment? I had a quick go just mocking the 1 function I used and it was trying to do the whole init() function and I got a bit lost after that aha.

davidzwa:
API endpoints should not return 500. Thats a code for bugs or things that the developer should fix on the Octofarm server itself. 
400 series is for bad input or problems.

NotExpectedYet:
> API endpoints should not return 500. Thats a code for bugs or things that the developer should fix on the Octofarm server itself.
> 400 series is for bad input or problems.

It is a bug server side if that fails

NotExpectedYet:
There's a whole class of codes for Sever error, which is what this is catching so not sure what your on about with "API's shouldn't not return 500". 

The 5xx (Server Error) class of status code indicates that the server is aware that it has erred or is incapable of performing the requested method

davidzwa:
Yeah in hindsight my comment was a bit weird. I could explain how I come to think of 5xx, but its subjective probably. Lets move on haha

NotExpectedYet:
@davidzwa sound aha if you have reasoning I'm happy to hear it but no drama either way. XD

NotExpectedYet:
Is up to you regarding the tests. It's good to go apart from that I believe.

There's no output from the additional function I wrote in state, and it grabs stuff from OctoPrint so gather that's a larger task to test?

server_src/routes/printers.js
davidzwa: I dont understand this comment as this is the backend we're looking at?

NotExpectedYet:
@davidzwa I've moved the refresh button to the side bar above save now. Was there some bugs you mentioned?

I fixed an issue that's happens when the port preference is null, it would incorrectly show the "Port not available" sign when there is no port preference. Found it because when you turn virtual printers on, the port preference is wiped on OctoPrint causing the refresh on OctoFarm to add a null in the port list. I don't see this been an issue in real world application, OctoFarm could not update the port preference if it detects a null from OctoPrint & OctoFarm has a saved value. I've not noticed OctoPrint wiping it in any other situation, or just magically forgetting like was mentioned in the discord. The preference get's saved in config.yaml and I tested an OctoPrint restart and it persisted.

client_src/js/lib/octofarm_client.js
```diff
@@ -0,0 +1,42 @@
...
+        "Content-Type": "application/json"
+      },
+      body: JSON.stringify(data)
+    });
+    if (post.status === 400) {
+      throw new Error(
+        `Malformed request! Status: ${post.status} - ${post.statusText}`
+      );
+    } else if (post.status === 503) {
+      throw new Error("Error contacting server, is it alive?");
```
davidzwa, line 35: Are we ready for these errors to be thrown? Javascript breaks and quits on the first errors it sees.

server_src/lib/dataFunctions/printerClean.js
davidzwa: Im not a fan of this code style as the function becomes indeterministic. Simpler functions have simpler bugs.
Usually I create a 'getById' and a 'list' variant to be 100% clear, also because this is the CRUD approach (get, list are separated). Can you point to the code which becomes much simpler, help me see the impact a bit?

Btw, of course sometimes ease of use is really nice. I fully understand that 😉. Just wanted to let you know that in such cases either the code which uses this function could benefit from unit tests to make absolutely sure the `id` parameter is not left null/false/empty/0/NaN. If I see those test, you would not have heard me ring a bell as I would know you'd tested it and your thus in full control. 
Those tests are probably high in count, so splitting it up is: 1) less impactful 2) really clear 3) less code to cover in tests

davidzwa: image
Changing the function to 2 types doesnt seem like it will impact the code.

server_src/routes/printers.js
```diff
@@ -125,25 +127,26 @@ router.get("/groups", ensureAuthenticated, async (req, res) => {
...
-      storage: printers[index].storage,
-      fileList: printers[index].fileList,
-      systemChecks: printers[index].systemChecks
-    };
-    res.send(returnPrinter);
+router.post("/updatePrinterSettings", ensureAuthenticated, async (req, res) => {
+  const id = req.body.i;
+  if (!id) {
+    logger.error("Printer Settings: No ID key was provided");
+    res.statusMessage = "No ID key was provided";
```
davidzwa, line 138: I dont think this statusMessage scales. Reason is that error response often is more than just a message. It could become a code, a context object, a message - more than just a string. So, I foresee a moment where we'd have to revert this in future and that's why it doesnt work for me.

`throw "message"`
`throw new Error("error message")`
`throw new ValidationError(validationErrorInput)`
`throw new OctoPrintError(octoprintParseError)`
... others

All of these dont fit in statusMessage string.

davidzwa: Oh btw, that does not stop this PR. It is just that I would not like to do this globally. That's all im sayin

davidzwa (changes requested):
So to summarize
- This PR is epic
- Youre doing such nice work!
- I get to learn so much right now, this is so handy for thinking about stuff for V2
- Merge solved
- Fixed the tests
- I applied the list/getById change
- I applied prettier to some client files (only partially, more to be done)
- the error throwing in client `octofarm.js` code => risk analysis (maybe test with some test errors server-side, see how it looks?) I think this should be postponed for more breathing room as client side is a risk factor.
- After settings have been updated we see 3 messages (maybe a bit over the top)
image
- After saving the settings, in case we had the **red** portPreference problem and tried to correct for it, the red color does not disappear although the situation is fixed.
- Saving/refreshing is quite slow. Are there concessions possible to speed this up? Could it be that state.js is more work than necessary? (We can follow-up with a PR ofc)
- think about statusMessage, see my feedback above and think if this scales for the idea you have in mind in case we need to add error context (f.e. API validation errors or OctoPrint remote API errors). Please communicate what your plans are 👍🏼

client_src/js/lib/octofarm_client.js
NotExpectedYet: I try catch everywhere I call that. And plan to do so for the rest I migrate over.  I'd love a global error handler but it's not something I'd say was worth our time pending V2. Or, could be done in another PR as I do want to migrate the client over to this file but the codes a reet spaghetti junction aha.

server_src/routes/printers.js
NotExpectedYet: I'm not using the status message for those eventualities there though? It's literally just saying "dodo, you need to provide a printer Id for this to work". Errors are caught below and didn't fire the status message? Can't remember. Anyway I think a global error handler should be tackled in another PR and update to use it. Would be good V2 prep work no?

client_src/js/lib/octofarm_client.js
NotExpectedYet: I gather I can move the try and catch inside this file and handle it there? We would have a global OctoFarm connection error handler then?

NotExpectedYet: Wouldn't need to repeat it around the rest of the files either. Let me know will update it.

NotExpectedYet:
> So to summarize
> 
> * This PR is epic
> * Youre doing such nice work!
> * I get to learn so much right now, this is so handy for thinking about stuff for V2
> * Merge solved
> * Fixed the tests
> * I applied the list/getById change
> * I applied prettier to some client files (only partially, more to be done)
> * the error throwing in client `octofarm.js` code => risk analysis (maybe test with some test errors server-side, see how it looks?) I think this should be postponed for more breathing room as client side is a risk factor.

It was tested but commented above, regarding moving the try/catch then we can deal with it in the file. I'm not depended on using throw new error, would just like to know alternatives and how better to deal with it. 

> * After settings have been updated we see 3 messages (maybe a bit over the top)
>   image

Those are because of the 3 calls the saving settings actually makes. It contacts OctoFarm, it contacts OctoPrint's API twice. I'll see about putting it all into a single notification and doing warning colour if one fails instead. 

> * After saving the settings, in case we had the **red** portPreference problem and tried to correct for it, the red color does not disappear although the situation is fixed.

Hmm that went on mine, will double check. Any reproduction? as I was using virtual printers for testing. 

> * Saving/refreshing is quite slow. Are there concessions possible to speed this up? Could it be that state.js is more work than necessary? (We can follow-up with a PR ofc)

Not from my experience here. Saving was rapid and so was refresh. Is this a windows vs linux thing again? 

> * think about statusMessage, see my feedback above and think if this scales for the idea you have in mind in case we need to add error context (f.e. API validation errors or OctoPrint remote API errors). Please communicate what your plans are, my opinion is flexible but my code-smell nose is not 👍🏼

Replied above.

client_src/js/lib/octofarm_client.js
NotExpectedYet: I always think of throw new Error, if caught to continue with the rest of the code. If this response is from the server then it's considered an error to me. Should this not be the case?

davidzwa: All Im saying is that I spotted new code with an explicit `throw new Error(...)` which if thrown could break the UI.
If you are certain that we try-catch everything, sure no prob. I dont oversee that.

Oh note that code 503 and 400 are rare right now (400 is rarely thrown because we dont have input validation everywhere, its why I proposed using `npm-input-validation` as shown in octoprint-companion PR). 
Codes 0, 500, 404 are more common and I dont see all of those covered.

Oh and why an else on 204 if it doesnt do anything but return?

server_src/routes/printers.js
davidzwa: If your intention is to return input validation result, I propose to use `npm-input-validator` and return that result **in the body** + **400 code**. Much shorter code-wise, much clearer for another developer and much easier for the client to recognize and work with. This abstraction comes with so much benefit!
Here's how the input validation rules can look:
```
const checkInputRules = {
  persistenceUuid: "required",
  deviceUuid: "required",
  host: "string",
  port: "required|numeric",
  docker: "required|boolean",
  allowCrossOrigin: "required|boolean"
};
```
and this is the response if I leave out properties on purpose:
```
{
    "deviceUuid": {
        "message": "The device uuid field is mandatory.",
        "rule": "required"
    },
    "port": {
        "message": "The port field is mandatory.",
        "rule": "required"
    },
    "docker": {
        "message": "The docker field is mandatory.",
        "rule": "required"
    }
}
```
I had to do nothing for it (except for calling the validation and returning it if failed)! And because the rules are constant strings, we dont have to guess what is the reason for failure on the client side. Its explicit and reliable.

If input validation is what you're looking to abstract, I hope you like the method I propose above. This would make porting API code over to V2 an absolute breeze! 💨 💨💨

davidzwa:
> > * the error throwing in client `octofarm.js` code => risk analysis (maybe test with some test errors server-side, see how it looks?) I think this should be postponed for more breathing room as client side is a risk factor.
> 
> It was tested but commented above, regarding moving the try/catch then we can deal with it in the file. I'm not depended on using throw new error, would just like to know alternatives and how better to deal with it.

Responded above

> Those are because of the 3 calls the saving settings actually makes. It contacts OctoFarm, it contacts OctoPrint's API twice. I'll see about putting it all into a single notification and doing warning colour if one fails instead.

That's nice

> Hmm that went on mine, will double check. Any reproduction? as I was using virtual printers for testing.

I used my Pi4 + prusa mini to reproduce as this is a portPref/udev problem and virtual does not work for that? This is a physical problem.
1) I turned on my printer (ttyACM0, not VIRTUAL)
2) connected to it in OP + saved
2b) checked config: ttyACM0 saved
3) powered off the printer 
4) I refreshed OctoPrint to check the 'usb serial not found blabla'
5) I checked OF, and went to the printer dialog (it was red: expected)
6) I rectified it in the dialog and pressed save.
7) Saw that the red color didnt dissappear. If Im correct I also pressed the refresh to make sure.
7b) I checked OP + config: problem was gone there and serial was set back to VIRTUAL.
8) I reopened the dialog and the red color was gone, so most probably just UI logic to change the color missing or not working as expected.

So I think this will occur for your situation as well. Please try again.

> > * Saving/refreshing is quite slow. Are there concessions possible to speed this up? Could it be that state.js is more work than necessary? (We can follow-up with a PR ofc)
> Not from my experience here. Saving was rapid and so was refresh. Is this a windows vs linux thing again?

My prusa/Pi4 is running on the network, so I think this is just delay because of calling multiple OP API endpoints in a row synchronously. My question is can we optimize this? If I save printer settings, I dont want the UI to wait for regenerating cache 
 or calling to reload profiles, since we have SSE for that (and hopefully WS later). Look the reason Im hammering on this is because it seems that this is suboptimal and the core business of OF to be as snappy as possible.

This has nothing to do with Windows. That was not a problem. The reason for the 5 sec time was because SystemRunner just does a lot of work for 

> 
> > * think about statusMessage, see my feedback above and think if this scales for the idea you have in mind in case we need to add error context (f.e. API validation errors or OctoPrint remote API errors). Please communicate what your plans are, my opinion is flexible but my code-smell nose is not 👍🏼
> 
> Replied above.

I hope you like the `npm-input-validator` approach, now that I showed some examples for it. This can be followed up in another PR if you wish. This work is the biggest gap to fill to hop onto the V2 train btw, especially because I see how the cleaners and state work now. Well worth our time to discuss

client_src/js/lib/octofarm_client.js
NotExpectedYet: There's an else on 204 because I use 204 as a confirmation response from the server where no return data is needed. The return the function to let the rest skip which is at the bottom converting anything else to a JSON object. There would be nothing to convert if it didn't return and stop the .JSON.

NotExpectedYet:
Ah, I didn't plan for that eventuality. I went to octoprint and resolved the issue then rescanned. I didn't sort in OctoFarm because I presumed they'd have to go there. The refresh cleared it for me though. I gather all that needs is a refresh of the page on save as that isn't done currently.

NotExpectedYet:
@davidzwa Made the changes, and I've removed most of those calls from state. One of the checks isn't firing for reScaning the OctoPrint instance when you updated the URL that time but think that function needs cleaning up in another PR. The whole difference function checks need looking at tbh. Anyway, now saves. I decided not to put the try/catch inside the function for octofarm_client as I think the response should be handled higher up. I've also captured a screen recording of exactly how fast the older method of refresh/saving was for me so I don't think it's the multiple calls to OP that's causing the lagged response on your end unless I'm misunderstanding you, to me that would be reproducible here no? 

Peek 2021-05-29 07-38

server_src/runners/state.js
```diff
@@ -3473,26 +3479,7 @@ class Runner {
         });
       }
 
-      await Runner.getProfile(settings.printer.index);
-      await Runner.getSettings(settings.printer.index);
-      await Runner.getOctoPrintSystenInfo(settings.printer.index);
-      Runner.getOctoPrintSystenInfo(settings.printer.index);
-      Runner.getUpdates(settings.printer.index);
-      Runner.getPluginList(settings.printer.index);
```
davidzwa, line 3481: Mark this in your memory that if there are problems with settings we can add any of it back. I dont think we will have to, but just note.

davidzwa (approved):
YAAAAS

PR #550, opened 2021-05-06
🔌 Octoprint companion plugin (OIDC) & plugin manager

davidzwa:
This PR focuses on adding the registration API assuming the current OctoPrint setup without HTTP/HTTPS tunneling, so the current way.

Input:
- `deviceKey` a key to indicate a pre-configured device ready to be paired to OctoFarm. This key is mandatory and serves as phase-1 UUID mechanism. It does not need to be unique compared to other OctoPrint instances.
- `persistenceKey` is required as a restart-safe key to deduplicate OctoPrint devices, which does not survive backups.
- `clientId` a pre-shared identifier for the OpenID `clients_credentials` auth flow (todo)
- `clientSecret` a pre-shared secret for the OpenID `clients_credentials` auth flow (todo)
- `octoPrintVersion` as a version safety mechanism (todo)
- `octoPrintPluginVersion` as a version safety mechanism (todo)

Some thoughts:
- OctoPrint plugin can push automatically by config, but then we need further rectification on OctoFarm side. So these printers are listed as 'setup required' or 'check required'.
- Users should be able to force the registration through OctoPrint UI as well. This gives them the ability to send the URL as well immediately finishing the registration without setup on OctoFarm.
- Both OctoPrint and OctoFarm do checks in the pairing process to check if the URL's given are valid and provide expected response
- ~We need a pairing key generator, but alternatively we could provide a QR code on OctoFarm which can be scanned by opening OctoPrint on mobile.~ We will configure a autogenerated client_id and client_secret.
- ~Can we assume FindMyOctoprint? I dont think so, unless we include the code in our plugin~ for later, or for a different stage of the octoprint plugin
- OpenID client credentials flow to authenticate both OF and OP

OctoPrint config:
- `device_key`: the OctoFarm API key stored in config
- `octofarm_host`: the OctoFarm base URL
- `octofarm_port`: the port for the OctoFarm server
- `client_id` a pre-shared identifier for the OpenID `clients_credentials` auth flow (todo)
- `client_secret` a pre-shared secret for the OpenID `clients_credentials` auth flow (todo)
- `ping`: plugin re-registration interval in secs (minimum 30)

Progression:
- [x] Add API endpoint/webhook registration
- [x] Finish API endpoint registration
- [x] OIDC server and db adapter
- [x] OIDC guard (introspection hack)
- [ ] Add printer registration panel

Problematic if not done before release (database nukes upcoming...)
- [ ] Rewrite Mongo and memory adapter to be normal Model adapter
- [ ] Make sure nodemon doesnt trigger restart when the printer announcement is made (OIDC causing this?)

Tests:
- [x] Plugin manager test
- [x] `octoprint-companion` plugin tests
- [x] Add input validation tests
- [x] Add valid registration test
- [ ] Add duplicate deviceKey, unique persistenceKey test
- [ ] Add duplicate deviceKey, duplicate persistenceKey and new IP/uuid test (we need to talk about this)


davidzwa:
https://gist.github.com/davidzwa/60c9b5ffd831059a1272ac9be92d6cb3

Plugin code

server_src/routes/plugins/octoprintRegistration.js
```diff
@@ -0,0 +1,165 @@
...
+    registrations
+  });
+});
+
+const registrationValidationRules = {
+  persistenceUuid: "required",
+  deviceUuid: "required",
+  deviceUrl: "required",
+  deviceApplicationKey: "required"
+};
```
davidzwa, line 109: @NotExpectedYet this is what I meant with input validation.

server_src/routes/plugins/octoprintRegistration.js
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * Convert an announced instance to a printer, requiring the 'deviceUrl' and 'deviceApplicationKey' to perform registration as normal.
+ */
+router.post("/register", ensureAuthenticated, async (req, res) => {
+  // First validate input
+  const validInput = await handleInputValidation(
```
davidzwa, line 115: @NotExpectedYet this is what I meant with input validation how I abstracted it.

server_src/routes/plugins/octoprintRegistration.js
```diff
@@ -0,0 +1,165 @@
...
+const { Runner } = require("../../runners/state");
+const { ensureAuthenticated } = require("../../config/auth");
+const { Validator } = require("node-input-validator");
+const router = express.Router();
+
+async function handleInputValidation(req, rules, res) {
+  const v = new Validator(req.body, rules);
+  const matched = await v.check();
+  if (!matched) {
+    res.status(422).send(v.errors);
```
davidzwa, line 13: @NotExpectedYet this is what I meant with returning a 4xx code on input validation. I must admit I'd rather use 400, so this is just copied over from the example I found.

server_src/routes/plugins/octoprintRegistration.js
```diff
@@ -0,0 +1,165 @@
...
+ * Convert an announced instance to a printer, requiring the 'deviceUrl' and 'deviceApplicationKey' to perform registration as normal.
+ */
+router.post("/register", ensureAuthenticated, async (req, res) => {
+  // First validate input
+  const validInput = await handleInputValidation(
+    req,
+    registrationValidationRules,
+    res
+  );
+  if (!validInput) return;
```
davidzwa, line 120: As the response is already sent, the code is already set, we just return.

server_src/routes/plugins/octoprintRegistration.js
NotExpectedYet: Aye I need to have a solid look at it, the package looks nice!

davidzwa:
@NotExpectedYet Im thinking of finishing this plugin in beta mode and disabling it by default as a setting. Its very near being done, but security is not up to scratch. 

I was thinking of building a notifications view for reviewing OP instances which announced themselves (and weren't registered previously). This way the user can verify that this is legit.
Only problem is that this announcement strategy is unauthenticated and can be eavesdropped due to the HTTP connection.

davidzwa:
Current state:
- [x] OIDC companion plugin provider
https://github.com/panva/node-oidc-provider
- [x] plugin manager + core scanning tests
- [ ] Core scanning fixed
- [ ] plugin manager - plugin configuration (wip)
- [ ] Fixed plugin configuration tests 
- [ ] plugin manager - plugin boot + API tests (future)
- [ ] OIDC client credentials flow tested locally 
- [ ] OIDC client credentials flow tested with postman
- [ ] registration views created
- [ ] registration views tested
- [ ] registration of printers works

This PR will not cover: plugin configuration, plugin views (we will not abstract this yet), plugin assets, plugin db migration scripts

davidzwa:
This PR is nearing completion. Cant believe Im actually saying it 😲

davidzwa:
Will re-open when ready.

PR #564, opened 2021-05-11
Chore/package updates

davidzwa:
{edit] Ready for merge. Should be harmless, but we should test.


NotExpectedYet (approved):
These are fine, and had no impact that I could tell. They're mainly dev dependencies anyway apart from .env.

davidzwa:
I will test this later again as well, gotta check gulp etc

NotExpectedYet:
> 
> 
> I will test this later again as well, gotta check gulp etc

It built fine for me like but no drama.

davidzwa:
@NotExpectedYet postcss wasnt installed for me as it was a peer-dep for gulp-postcss... pretty weird must be a bug on their side (I guess windows only or maybe a transient installation problem)

NotExpectedYet:
> 
> 
> @NotExpectedYet postcss wasnt installed for me as it was a peer-dep for gulp-postcss... pretty weird must be a bug on their side (I guess windows only or maybe a transient installation problem)

I tested on Windows but gather was already in my node_modules as I hadn't cleared before. Yeah weird one but good catch.

PR #568, opened 2021-05-12
Bugfix/567 heatmap race condition

davidzwa:
- [x] Add heatmap tests + repro 
- [x] Add farm info service
- [x] Fix startup index/app-core race condition **OR** patch the heatmap on the spot
- [x] Changelog

Unrelated but small changes:
- [x] Generic info to debug log changes
- [x] `undefined` to "undefined" change (although not the cause of problems thusfar)
- [x] Revert `filamentManager` renaming `filamentManagerEnabled` to avoid future headaches (either rename it all, or leave it untouched)


server_src/settings/serverSettings.js
```diff
@@ -108,5 +108,5 @@ class ServerSettings {
 
 module.exports = {
   ServerSettings,
-  filamentManagerEnabled: filamentManager
+  filamentManager
```
davidzwa, line 111: Me reverting my half-assed attempt at renaming a setting. I only foresee issues if this is not done all the way thru the UI, database etc. Thats why I reverted it.

server_src/lib/tests/printer-clean.test.js
```diff
@@ -0,0 +1,79 @@
...
+      farmStart: new Date(),
+      heatMap: [
+        {
+          name: "Completed",
+          data: []
+        }
+      ]
+    }])
+
+    expect(PrinterClean.initFarmInformation()).rejects.toThrow("Cannot read property 'x' of undefined");
```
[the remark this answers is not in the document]
NotExpectedYet: We could define the heat map in the database no? We'd only have to check for the day in the code if I remember rightly.

server_src/settings/serverSettings.js
NotExpectedYet: Yeah that's a lot of places xD

server_src/lib/dataFunctions/printerClean.js
```diff
@@ -1548,48 +1548,21 @@ class PrinterClean {
...
-      newfarmStats.save();
+
+      farmStats[0] = await FarmStatisticsService.create(farmStart, heatMap);
     } else if (typeof farmStats[0].heatMap === "undefined") {
       farmStats[0].heatMap = heatMap;
       dashboardStatistics.utilisationGraph = heatMap;
       farmStats[0].markModified("heatMap");
-      farmStats[0].save();
-    } else {
-      heatMap = farmStats[0].heatMap;
```
davidzwa, line 1566: Deleted it @NotExpectedYet as it is duplicate, up to you to tell if this is creating problems down the line.
At least my test didnt detect a throw anymore, so the bug is gone.

server_src/lib/tests/printer-clean.test.js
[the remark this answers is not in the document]
davidzwa: See my change below, I think you're right but I saw incorrect duplication so maybe we can prevent this bug differently.

server_src/lib/dataFunctions/printerClean.js
NotExpectedYet: I think your right here. That extra save was probably an early attempt to fix another issue.

davidzwa: Lol that makes me feel really uneasy

NotExpectedYet: I plan too, I wasn't meaning I wasn't going to haha

NotExpectedYet:
I'll get this checked tomorrow and go through the file/job clean one once merged.

NotExpectedYet (approved):
Tested and all seems working from this end. 

- [x] Existing Mongo Local on PC issue was discovered: Result - Original database recovered. 
- [x] New Mongo Local on PC issue was discovered: Result - New database generated 
- [x] Existing mongo db on Atlas: Result - Original database recovered
- [x] Remote mongo db on Atlas: Result - New Database Generated

Good spot!

PR #575, opened 2021-05-14
574 state split

davidzwa:
- [x] Split off OctoPrint API base service (more enhancement to be added later)
- [x] Implement OctoPrint API service mock
- [x] Add Runner.init test
- [x] Test key regression (global api key)

Any other stuff
- [x] Discuss and decide on timeout state variable
- [x] Bugs fixed
- [x] Changelog

### DIFF
--- a/server_src/runners/state.js
+++ b/server_src/runners/state.js
@@ -1954,9 +1954,7 @@ class Runner {
         );
         const { _id } = farmPrinters[index];
         await this.setupWebSocket(_id, skipAPI);
-      } else if (
-        farmPrinters[index].ws.instance.readyState === 2
-      ) {
+      } else if (farmPrinters[index].ws.instance.readyState === 2) {
         PrinterTicker.addIssue(
           new Date(),
           farmPrinters[index].printerURL,
@@ -2454,7 +2452,7 @@ class Runner {
       "Active",
       farmPrinters[index]._id
     );
-    
+
     return this.octoPrintService
       .getPluginManager(farmPrinters[index], true)
       .then((res) => {
--- a/server_src/runners/state.js
+++ b/server_src/runners/state.js
@@ -1,6 +1,7 @@
+"use strict";
+
 const _ = require("lodash");
 const EventEmitter = require("events");
-const fetch = require("node-fetch");
 const WebSocket = require("ws");
 const Logger = require("../lib/logger.js");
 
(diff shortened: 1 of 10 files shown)


davidzwa:
@NotExpectedYet this PR is ready for review, but not ready for merge (I will only add tests from this point)

server_src/runners/state.js
```diff
@@ -977,85 +979,17 @@ WebSocketClient.prototype.onclose = function (e) {
...
-}
-
 class Runner {
+  static octoPrintService = undefined;
+
   static async init() {
     farmPrinters = [];
     const server = await ServerSettings.check();
     systemSettings = server[0];
     timeout = systemSettings.timeout;
```
davidzwa, line 989: This is the timeout variable I mean. The `OctoprintApiClientService` has a copy of it. Its shared a couple lines below.

NotExpectedYet:
@davidzwa I'm not sure what you mean as there's a websocket timeout, and two API timeouts. They're seperate. I'll take a deeper look in the morning anyhu aha. Refresh my memory.

server_src/runners/state.js
NotExpectedYet: @davidzwa that's the object from the database. Timeout.apiretry, timeout.websocket yada. Check the server settings modal. I put in a variable for global access at the time due to not using a cache.

server_src/services/octoprint/octoprint-api-client.service.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  async getPrinterProfiles(printer, retry = false) {
+    return this.getWithOptionalRetry(printer, apiPrinterProfiles, retry);
+  }
+
+  async getPluginManager(printer, retry = false) {
+    const printerManagerApiCompatible = checkPluginManagerAPIDeprecation(printer.octoPrintVersion);
+
+    const route = printerManagerApiCompatible ? apiPluginManagerRepository1_6_0 : apiPluginManager;
```
davidzwa, line 91: I love this kind of coding, really clean short functions with seasoning in the right places

davidzwa:
@NotExpectedYet this works beautifully for me. Would be nice if this hits dev soon because its really worth it

NotExpectedYet:
This is looking good man, will study it more today. Wanna make sure I understand it all aha.

davidzwa:
@NotExpectedYet nice, I will give you your time.

The code is just an abstraction over your implementation. So timeout and retry mechanism is one function instead of copied each time.
The base class OctoprintApiService provides all the GET, POST, PATCH, etc. The OctoprintClientApiService class extends on that and implements specific API functions to call.
Mocked only the base service for tests, worked a treat so far. The mock is simple: you set the expected response first and then you call the service. You can then expect the response or failures. Pretty neat

NotExpectedYet (approved):
Tested and reviewed all looks good to me. Still reacts to timeout changes from what I can tell. Good to go! Websockets next xD. Then I can start looking at Klipper / Marina first.

Issue #542, opened 2021-05-03
Deprecation warning for "ensureIndex"


NotExpectedYet:
### Description of the problem

```bash
(node:118383) DeprecationWarning: collection.ensureIndex is deprecated. Use createIndexes instead.
```

### Reproduction of the problem

```bash
Start the app
```

PR #580, opened 2021-05-15
Upgrade mongoose, fix circular import, fix ensureIndex warning

davidzwa:
- [x] #542 ensure index warning avoided
- [x] Update mongoose
- [x] Remove circular Runner import
- [x] Changelog


NotExpectedYet (approved):
Sweet! Ah so that's where the circular bug was!

PR #582, opened 2021-05-16
🐦 canary 457 settings refactor for System Settings

NotExpectedYet:
Initial Settings Refactor for System. 

Some warnings with this PR. It's quite big due to it updating the rest of the usages of the settings so I could finally clear out those two files. 
It is not! a refactor of code where settings are used. Mainly history / state.js files. Those are to be tackled in a new PR. Same with the API endpoints. I took the stand point to try and keep this small that once these functions are replaced (to keep functionality), the API/index/history can be updated at a later date.

 - [x] Create a new octofarm service specifically for system settings
 - [x] Create a new cache for system settings
 - [x] Initialise settings on server boot and create a cache
 - [x] Initialise middleware for user/group and utilise on app.use()
 - [x] Allow for database to be completely reset
 - [x] Allow for database key/value pair to be updated
 - [x] Allow for retrieval of specific database key/value pair
 - [x] Setup new middleware function for EJS templates to always load system settings
 - [x] Setup helper functions for the current database system.
 - [x] Replace existing functions within the system
 - [x] convert to class to appease the v2 gods
 - [x] update cache to instantiate the class and store
 - [x] re-update the rest of the changes to use the cache
 - [x] move middleware into 'middleware' folder
 - [x] Fix current tests 
 - [ ] Replace api settings grabs
 - [ ] Replace api updates
 - [ ] Validation on settings API endpoints
 - [ ] JSDocs
 - Not related ~Restart of server doesn't account for not been logged in~
 - ~Look up how to mock the class and get cache function~ not needed, cache is already in memory and can be tested directly
 
Tests
- [ ] Jest function tests
- [ ] Test settings server loginRequired false, true and unitialized
- [ ] Test MVC routes filament, index, printers, serverChecks, settings, SSE-monitoring, users
- [ ] Fix settings getters being called as functions (they are properties, not functions!)
- [ ] Remove unnecessary server settings cache mock
- [ ] Rename middleware and settings init functions app-core to make more sense


NotExpectedYet:
hmm da fuck did I break XD

davidzwa:
The test action failing is what I meant. Gotta fix that asap...

davidzwa:
@NotExpectedYet if you can guide me to what help you need here, I will step in to do so before picking up anything else.

NotExpectedYet:
@davidzwa 

So I gather we need to create a mock class for the settings cache and the settings service. 

That can be utilised in the history tests you did where it calls for the filament manager settings. 

The docs I'm fine with, will sort those. 

Once that's done I'd guess we could run tests on the API end points but up to you whether we add them in this PR or another as it's already growing.

I plan to add validation in the cache function as it seems an apt place to put it, we can also use mongoose for that. It isn't too hard to add into the current function set for findOneAndUpdate. I just need to make the current settings service stuff react to the rejection and remove the initial validation. 

The settings endpoint just pretty much needs the function calls updating, which I'm part way through. Should get time to finish at least that off by tomorrow.

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,294 @@
...
+    // Find the filter and update the server settings with the new value...
+    const updatedSettings = await ServerSettingsDB.findOneAndUpdate(
+      { _id: filter },
+      { server: value },
+      {
+        new: true
+      }
+    );
+    // Make sure to replicate the changes in cache.
+    this.currentServerSettings["server"] = value;
```
davidzwa, line 124: A string prevents refactored renaming in an IDE like WebStorm. In general I'm not fond of string based code, although in Python we see it a lot sadly. What is your reasoning to do so here and beyond?

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,294 @@
...
+  get entireSettingsObject() {
+    return this.currentServerSettings;
+  }
+
+  /**
+   * Returns the polling settings for OctoPrints websocket connection
+   * @returns {Object} Object from cache for OctoPrints websocket connection
+   */
+  get octoPrintWebsocketPollingSettings() {
+    return this.currentServerSettings["onlinePolling"];
```
davidzwa, line 65: Can you explain to me the reason for these getter functions? I can think of many, just interested in hearing your thoughts on it, to be able to check that this stays manageable when it scales up

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,294 @@
...
+  }
+  /**
+   * Updates the polling settings for OctoPrints websocket connection in the database and then the cache.
+   * @param {Object} value containing the seconds key for OctoPrints websocket polling. String value of polling time.
+   * @throws {Error} value not provided or empty.
+   * @returns {Object} containing the new settings key
+   */
+  async updateOctoPrintWebsocketPollingSettings(value = {}) {
```
davidzwa, line 73: Nice to see the explicitness in your function naming. This is maybe a bit long tho, eager to hear your thought as in C# Im constantly struggling with it

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,294 @@
...
+   * @returns {Object} Object from cache for OctoPrints websocket connection
+   */
+  get octoPrintWebsocketPollingSettings() {
+    return this.currentServerSettings["onlinePolling"];
+  }
+  /**
+   * Updates the polling settings for OctoPrints websocket connection in the database and then the cache.
+   * @param {Object} value containing the seconds key for OctoPrints websocket polling. String value of polling time.
+   * @throws {Error} value not provided or empty.
+   * @returns {Object} containing the new settings key
```
davidzwa, line 71: The return type is not a key, but a value ;) Thats quite different - it confused me when reading this jsdoc. Thought I'd mention it sincerely

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,294 @@
...
+   */
+  async updateOctoPrintWebsocketPollingSettings(value = {}) {
+    const model = new ServerSettingsDB(value);
```
davidzwa, line 74: This is not a model. Its a document, row or entity. A model is the code/representation.
Found it important to mention as lingo is truly a handy way to quickly communicate code

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,294 @@
...
+
+    const validatedModel = model.validateSync();
+    if (!!validatedModel) throw validatedModel;
```
davidzwa, line 77: If this is what you'd like and I support it, try to create 1 function for it so we dont have to copy this code for other sub-settings (children in the settings document)

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,294 @@
...
+
+    //Filter out the current settings from cache
+    const filter = this.currentServerSettings._id;
```
davidzwa, line 80: The variable is a document ID which I'd name `settingsId` or similar, a filter tells me its an object with multiple keys/values.

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,294 @@
...
+    await currentDatabaseSettings.save();
+    this.currentServerSettings = currentDatabaseSettings;
+  }
+  /**
+   * Checks the database for the server settings, can be supplied with filter and options to specify mongoose options/filter.
+   * @param {Object} filter object to filter the database results
+   * @param {Object} options object to change the database options
+   * @returns {Object} Object containing the filtered settings / full settings
+   */
+  returnDatabaseValues({ filter = {}, options = {} } = {}) {
```
davidzwa, line 49: Function naming unclear to me, suggestion `listServerSettingsDocs`. Oh and btw, I think we're aiming to have 1  server settings, so maybe change it to `getSettings`. Usage would then be `this.getSettings` which is clear to the internal methods here as well to the outside which use the ServerSettings service/repository.

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,294 @@
+const ServerSettingsDB = require("../../models/ServerSettings.js");
+
+const { isEmpty } = require("lodash");
+
+class ServerSettings {
```
davidzwa, line 5: Personally I'd call this `ServerSettingRepository` or `ServerSettingService` to be able to discern it from the database level (Schema's, low-level) as well as API level (REST validation, high level).  

To explain the difference I have in mind:
1) The Repository is a tool to wrap/extend/abstract/secure **storage** sources like file or database
2) the Service is a tool to combine underlying **data sources** so file/db/cache/middleware/external-api's/etc. The service is higher up and will do less low-level work like f.e. checking whether the values provided match our constraints (validation). It has little to do with the storage level (abstraction).

Therefore when I would've coded this, I'd call this a ServerSettingsService - it combines a cache with another storage (db). Secondly, I would move certain functionality to the cache to make it more useful and I would've created a ServerSettingsRepository class to extend the database work like explicit validation in order to make it much easier to get rid of MongoDB dependency (decoupling).

Let me know if you like this idea. I can also work it out in drawings to make it clearer.

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,294 @@
...
+
+    //Filter out the current settings from cache
+    const filter = this.currentServerSettings._id;
+
+    // Find the filter and update the server settings with the new value...
+    const updatedSettings = await ServerSettingsDB.findOneAndUpdate(
+      { _id: filter },
+      { server: value },
+      {
+        new: true
```
davidzwa, line 120: What is the reason for this?

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,294 @@
...
+    // Find the filter and update the server settings with the new value...
+    const updatedSettings = await ServerSettingsDB.findOneAndUpdate(
+      { _id: filter },
+      { influxExport: value },
+      {
+        new: true
+      }
+    );
+    // Make sure to replicate the changes in cache.
+    this.currentServerSettings["influxExport"] = value;
```
davidzwa, line 257: To summarize my feedback: lot of repetition in this file, in general a good direction but it can be improved. Try to be wary when you find yourself copying code structures.

server_src/services/database/server-settings.service.js
davidzwa: Oh btw, V2 and .NET based servers seem to do this pattern. 
- (Module) - hierarchical separation of code (registers Controller, Service, Repository, Entity and others)
- Controller - API and MVC request/response actions
- Service - Data source functionality (used by Controller)
- Repository - Data storage functionality (used by Service)
- Entity - Data modeling (used by Repository)

Some other constructs: 
- Manager - a entity manager responsible for 1 or multiple entities at the same time (used by Repository)
- Cache - special kind of storage controller in-memory
- Config - compact configuration model and manager (bit special as configuration is usually a startup flow where are merge action happens from terminal
- Constants/Data - constant data which define our application 

Note that you've introduced a PatreonsCache, but it contains no logic. I'd name that file PatreonsData or add it to the file with AppConstants

davidzwa:
> So I gather we need to create a mock class for the settings cache and the settings service.

I dont think we need to. We have a in-memory database during tests which works brilliantly. The cache is already in-memory. I'd skip mocking it.
 
> That can be utilised in the history tests you did where it calls for the filament manager settings.

Same reasoning

> The docs I'm fine with, will sort those.

Coolio

> Once that's done I'd guess we could run tests on the API end points but up to you whether we add them in this PR or another as it's already growing.

Absolutely, test it all. No API endpoint left untouched

> I plan to add validation in the service functions as it seems an apt place to put it, we can also use mongoose for that. It isn't too hard to add into the current function set for findOneAndUpdate. I just need to make the current settings service stuff react to the rejection and remove the initial validation.

I'd try to stay away from database level validators as we'd couple ourselves to the database more tightly. In code we have the control, so I'd prefer it to be in our hands. I do like the mongoose validators and I would be lying if I dont constrain my databases as well but I use that as an end-stop failure prevention. In the end I like to rely on validation we can call for cache/services/database and mongoose is only for the database.
In V2 we have the options to use the package `class-validators` which uses Typescript loveliness, but for V1 `node-input-validator` or `validator.js` are good non-es6 alternatives.

> The settings endpoint just pretty much needs the function calls updating, which I'm part way through. Should get time to finish at least that off by tomorrow.

Okido, sounds cool. Some tests fail currently because `req.isAuthenticated()` is returning true in config/auth.js `ensureAuthenticated` during tests but no login was provided. I dont understand why.

server_src/cache/patreon.cache.js
```diff
@@ -0,0 +1,28 @@
+// Snuck in here, no use yet...
+module.exports = [
```
davidzwa, line 2: Not really a cache, just data or constants

server_src/cache/__mocks__/server-settings.cache.js
```diff
@@ -0,0 +1,23 @@
+let { getServerSettingsCache } = jest.fn("../server-settings.cache.js");
+jest.mock("../../services/database/server-settings.service.js");
+let {
+  ServerSettingsMock
+} = require("../../services/database/server-settings.service.js");
+
+let mockedServerSettingsCache;
+
+getServerSettingsCache = async () => {
```
davidzwa, line 9: This mock is not really needed as we have an in-memory database

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,311 @@
...
+const { isEmpty } = require("lodash");
+
+class ServerSettings {
+  currentServerSettings;
+
+  /**
+   * Returns the entire settings object from catch
+   * @returns {Object} Object from cache for all of OctoFarms settings
+   */
+  get entireSettingsObject() {
```
davidzwa, line 12: Naming is weird to me. I'd name it `getAllSettingsGroups` (Groups, Sections, Children, whatever) to indicate that the settings are grouped and require traversing to get a specific setting

server_src/services/database/server-settings.service.js
```diff
@@ -0,0 +1,311 @@
...
+    return this.currentServerSettings;
+  }
+
+  /**
+   * Returns the polling settings for OctoPrints websocket connection
+   * @returns {Object} Object from cache for OctoPrints websocket connection
+   */
+  get octoPrintWebsocketPollingSettings() {
+    return this.currentServerSettings["onlinePolling"];
```
davidzwa, line 21: As mentioned before, do explain the thought behind the string based accessor?

app-core.js
```diff
@@ -55,20 +62,21 @@ function setupExpressServer() {
 async function ensureSystemSettingsInitiated() {
   logger.info("Checking Server Settings...");
 
-  await ServerSettingsDB.find({}).catch((e) => {
-    if (e.message.includes("command find requires authentication")) {
-      throw "Database authentication failed.";
```
davidzwa, line 60: This code was essential to trigger database startup issues. I dont think removing this is smart

app-core.js
```diff
@@ -55,20 +62,21 @@ function setupExpressServer() {
...
-    } else {
-      throw "Database connection failed.";
-    }
-  });
-
   // Setup Settings as connection is established
-  const serverSettingsStatus = await ServerSettings.init();
-  await ClientSettings.init();
-  logger.info(serverSettingsStatus);
+  const serverSettingsInitialisation = await initializeServerSettingsCache();
```
davidzwa, line 66: The lines of code in this function do not add a lot. I'd move the logger to within the cache initialization.

davidzwa:
image
Missing closing symbol, system.ejs line 416

server_src/cache/patreon.cache.js
davidzwa: Changed filename to make more sense. Up to you to consume it

davidzwa:
I advise to change `ensureCurrentUserAndGroup` to be aware of unitialized server settings. If undefined or falsy, it should not suddenly trigger a false positive scenario.
I would expect if system settings is undefined to throw an error and assume login is required.

server_src/services/database/server-settings.service.js
NotExpectedYet: I don't mind but made Polling, poll... shaved 3 characters!

server_src/services/database/server-settings.service.js
NotExpectedYet: These are still to be done, some are copied

server_src/services/database/server-settings.service.js
NotExpectedYet: I'm testing the validation, it's not staying like that

server_src/services/database/server-settings.service.js
NotExpectedYet: A singleton isn't a value though? To me I'm getting the database array of objects.

server_src/services/database/server-settings.service.js
NotExpectedYet: I was already planning too name it this aha, I do clean up last. Maybe I should save you typing on a draft pr and do clean up first XD. 

The patreons cache isn't technically anything to do with the application to me. It's a separate list. I'm thinking of putting it on the OF website and making OF call that json list from there. I'd just write the file locally so the EJS has something to dump on settings. I want a fallback though for air gapped farms.

server_src/services/database/server-settings.service.js
NotExpectedYet: > What is the reason for this `new` key and `true` value? Im both concerned as well as interested in what it does.

`New: true` tells mongoose to return the new object, not the original object which is it's default. You can also do `returnOriginal: false` but new was less characters and they are functionally the same. `returnOriginal` only exists for consistency with the actual nodejs mongo driver.

server_src/services/database/server-settings.service.js
NotExpectedYet: Yeah I know, but I'm personally leaning more towards creating explicit functions to handle these inside the code, rather than having to know the database values before hand and passing them through. 

I could create a function for updating the cache, but It was extra code to me vs repetition. 

So: 
```this.currentServerSettings.influxExport = value;```
VS:
``` this.updateServerSettingsCachce("influxExport", value);```
```updateServerSettingsCahce(key, value){ this.currentServerSettings[key] = value; }```

Is there another way?

**Comment by NotExpectedYet** [comment]
> > So I gather we need to create a mock class for the settings cache and the settings service.
> 
> I dont think we need to. We have a in-memory database during tests which works brilliantly. The cache is already in-memory. I'd skip mocking it. Personally I dont like the mocking as it is a waste of our keyboard typing time, but we have to sometimes to prevent stuff like file operation, external API calls and other weird edge cases.
> 

Colour me in with the had no idea pen, I'll study what you do xD. 

> I'd try to stay away from database level validators as we'd couple ourselves to the database more tightly. In code we have the control, so I'd prefer it to be in our hands. I do like the mongoose validators and I would be lying if I dont constrain my databases as well but I use that as an end-stop failure prevention. In the end I like to rely on validation we can call for cache/services/database and mongoose is only for the database.
> In V2 we have the options to use the package `class-validators` which uses Typescript loveliness, but for V1 `node-input-validator` or `validator.js` are good non-es6 alternatives.

We use another database it's gonna need another service file anyway so I dunno if it's worth, not utilising the validators in mongoose now. I get what your saying, but we can always bring in another validator for the new database service so I wouldn't say we we're entirely coupled to it, only coupled in the relevance to mongo which to me is fine. 

> Okido, sounds cool. Some tests fail currently because `req.isAuthenticated()` is returning true in config/auth.js `ensureAuthenticated` during tests but no login was provided. I dont understand why.
> This happens for `api/server-update.test.js` and `api/dashboard.test.js`

Did you sort this? Mad week/end just catching up aha. Remember speaking with you about it.

File: server_src/lib/tests/history-clean.test.js [hunk 18]
```diff
@@ -68,7 +74,7 @@ describe("historyClean", function () {
   it("should initiate and finish within 5 sec for empty history", async function () {
     expect(await mockHistoryService.find({})).toHaveLength(0);
 
-    const historyState = new HistoryClean(false, "info");
+    const historyState = new HistoryClean(true, "info");
```
davidzwa, line 77: It would be better not to write to file during tests. Please revert.

davidzwa:
> > I dont think we need to. We have a in-memory database during tests which works brilliantly. The cache is already in-memory. I'd skip mocking it. Personally I dont like the mocking as it is a waste of our keyboard typing time, but we have to sometimes to prevent stuff like file operation, external API calls and other weird edge cases.
> 
> Colour me in with the had no idea pen, I'll study what you do xD.

Please do.

> > I'd try to stay away from database level validators as we'd couple ourselves to the database more tightly. In code we have the control, so I'd prefer it to be in our hands. I do like the mongoose validators and I would be lying if I dont constrain my databases as well but I use that as an end-stop failure prevention. In the end I like to rely on validation we can call for cache/services/database and mongoose is only for the database.
> > In V2 we have the options to use the package `class-validators` which uses Typescript loveliness, but for V1 `node-input-validator` or `validator.js` are good non-es6 alternatives.
> 
> We use another database it's gonna need another service file anyway so I dunno if it's worth, not, utilising the validators in mongoose now. I get what your saying, but we can always bring in another validator for the new database service so I wouldn't say we we're entirely coupled to it, only coupled in the relevance to mongo which to me is fine.

Not saying we should consider compatibility with other databases. Im trying to say that we need more validation for cache input, API input, database input and sometimes even OP API response compatibility checks (Im really going far here in my mind). So mongoose validators only covers 1 out of 4. My point is therefore: we gotta do more work than just entity column/document validation and t.b.h. rather sooner than later. 

> > Okido, sounds cool. Some tests fail currently because `req.isAuthenticated()` is returning true in config/auth.js `ensureAuthenticated` during tests but no login was provided. I dont understand why.
> > This happens for `api/server-update.test.js` and `api/dashboard.test.js`
> 
> Did you sort this? Mad week/end just catching up aha. Remember speaking with you about it.

Yes, already applied here and you touched it with the switch statement. The problem was the false-positive scenario and the cause was that you queried `req.serverSettingsCache.loginRequired` instead of `req.serverSettingsCache.server.loginRequired`.

NotExpectedYet:
Force push at your peril, I'm currently working on this will merge locally.

davidzwa:
> Restart of server doesn't account for not been logged in

Does not fit in this PR. This is another thing which can be fixed with the Session Store. Not much to do with settings and one less thing for the check list ;)

NotExpectedYet:
@davidzwa agreed.

davidzwa:
@NotExpectedYet sorry dude this refactor is not compatible anymore with my latest work of the past days using `awilix` dependency injection. I "hope" to have that ready this week. I went over the code on dev so fast... this PR is not going to make it. Luckily I have very good reasons for it, and I think you'll completely agree with me once we've covered it all 😄 

I hope to show you how cool it will be, because `awilix` is awesome so far: everything falls into place... OF bootup, making sure stuff is initiated... testing, its truly crazy! (Also it feels much more like second nature to me...) Would be a shame to have to merge or delay with this PR because its older or long-awaited. There's definitely value here, but as I think something fundamental is going wrong (backend only! Client side I have no idea)

Its best to discuss the details and our strategy in DM!

NotExpectedYet:
I'll clean this up tomorrow as well and get it submitted to canary branch.

davidzwa:
#### Salvagable from this PR
- Client changes (with review pls)
- EJS changes (with review pls)
- Patreon data

### To be removed
- The middleware functions: totally superseded with upcoming work
- Settings route (deletion and changes in logic)
- System route (multer change and new endpoints)
- SSE changes (will be DI class)
- Filament route (will become DI controller... completely different approach + much easier to work with)
- State, History, Users, System, Settings, ServerChecks, Index, printers routes (=> controller)
- app, app-test, app-core changes concerning settings
- Settings Clean => absorbed in SettingsStore
- Client and Server Settings Cache (both)  => This has been stored in SettingsStore
- Influx => changed to DI service (so non-static, constructor injection)
- Test mocks (these were unnecessary)

davidzwa:
https://github.com/davidzwa/OctoFarm/tree/bugfix/457-settings-refactor
Backup pushed there

PR #585, opened 2021-05-17
Automatically switch to wss for https printer urls

davidzwa:
- [x] Switch to ws/wss protocols based on http/https
- [x] Tested for ws locally (in production)
- [x] Create utility for url conversion to socket
- [x] Add in Printer key for database storage
- [x] setDefaults should create the new websocket key if doesn't exist
- [x] Pass through the webSocketURL instead of the printer URL to the websocket 
- [x] Allow options in settings to change this and pass back up with saving
- [x] Changelog


NotExpectedYet (commented):
Sweet forgot about this.

davidzwa:
> Sweet forgot about this.

Might be nice to test this on the test instance you've enabled in the cloud.

NotExpectedYet:
@davidzwa hmm good shout suppose the octoprint instances would be behind a reverse proxy

davidzwa:
> @davidzwa hmm good shout suppose the octoprint instances would be behind a reverse proxy

I think `https://` does not always mean `wss://` is used, although I would hope so.

NotExpectedYet:
> > @davidzwa hmm good shout suppose the octoprint instances would be behind a reverse proxy
> 
> I think `https://` does not always mean `wss://` is used, although I would hope so.

Yeah this is why I never implemented it initially I couldn't find a solid answer on whether that's the case.

davidzwa:
Shall we abandon this or take the leap? I think a possible solution is a dropdown/checkbox to let the user fix the situation if our https=>wss choice was incorrect.

NotExpectedYet:
@davidzwa let me try it with a reversed proxy setup and see what happens. What my assumption was. I think that if you shove a reverse proxy that, when you enable websocket support does the wss -> ws route. Similarly if you enable a cert on the local device it switches the end point to wss. 

I don't think the option hurts at all incase that is completely wrong. I can jump in and enable that no worries.

NotExpectedYet:
@davidzwa  ready for a review from you

client_src/js/lib/modules/printerSettings.js
```diff
@@ -220,6 +220,11 @@ class PrinterSettings {
       currentPrinter.printerName;
     document.getElementById("psPrinterURL").placeholder =
       currentPrinter.printerURL;
+    // Convert websocket url into a url object
+    const webSocketURL = new URL(currentPrinter.webSocketURL);
+    // Grab out the protocol and select it on the select box.
+    document.getElementById("psWebSocketProtocol").value =
+      webSocketURL.protocol + "//";
```
NotExpectedYet, line 227: Added this into the client so we can grab the current protocol and select it on the select dropdown.

server_2x/src/api/legacy_runners/progress/state.js
```diff
@@ -1206,7 +1206,7 @@ class Runner {
                         farmPrinters[i]._id,
                     );
                     await farmPrinters[i].ws.open(
-                        `ws://${farmPrinters[i].printerURL}/sockjs/websocket`,
+                        `${farmPrinters[i].printerURL}/sockjs/websocket`,
```
NotExpectedYet, line 1209: Hmm thought this haddn't updated but it's your legacy runner aha. Couldn't you just import these rather than keeping a carbon copy update? No idea like thought I'd ask aaha

server_src/lib/dataFunctions/printerClean.js
```diff
@@ -375,6 +375,7 @@ class PrinterClean {
         group: farmPrinter.group,
         groups: [], // TODO unfinished feature
         printerURL: farmPrinter.printerURL,
+        webSocketURL: farmPrinter.webSocketURL,
```
NotExpectedYet, line 378: Make sure client get's the new key/value.

server_src/models/Printer.js
```diff
@@ -52,6 +52,10 @@ const PrinterSchema = new mongoose.Schema({
     type: String,
     required: false
   },
+  webSocketURL: {
+    type: String,
+    required: false
+  },
```
NotExpectedYet, line 58: stroregy in the databasey

server_src/runners/state.js
```diff
@@ -20,7 +21,9 @@ const { JobClean } = require("../lib/dataFunctions/jobClean.js");
 const { FileClean } = require("../lib/dataFunctions/fileClean.js");
 const { FilamentClean } = require("../lib/dataFunctions/filamentClean.js");
 const { PrinterTicker } = require("./printerTicker.js");
-const {checkPluginManagerAPIDeprecation} = require("../utils/compatibility.utils");
+const {
+  checkPluginManagerAPIDeprecation
+} = require("../utils/compatibility.utils");
```
NotExpectedYet, line 26: Honestly probably not going to be utilised anywhere else but rather than making state.js any bigger XD

server_src/runners/state.js
```diff
@@ -118,15 +121,8 @@ const heartBeatInterval = setInterval(function ping() {
 
 WebSocketClient.prototype.open = function (url, index) {
   try {
-    if (url.includes("http://")) {
-      url = url.replace("http://", "");
-    }
-    if (url.includes("https://")) {
-      url = url.replace("https://", "");
-    }
```
NotExpectedYet, line 126: Can remove this now as it's passed through with the variable

server_src/runners/state.js
```diff
@@ -1242,7 +1238,7 @@ class Runner {
             farmPrinters[i]._id
           );
           await farmPrinters[i].ws.open(
-            `ws://${farmPrinters[i].printerURL}/sockjs/websocket`,
+            `${farmPrinters[i].webSocketURL}/sockjs/websocket`,
```
NotExpectedYet, line 1241: Updated to open with the new variable

server_src/runners/state.js
```diff
@@ -1655,6 +1651,11 @@ class Runner {
         ""
       );
     }
+    if (!farmPrinters[i].webSocketURL) {
+      farmPrinters[i].webSocketURL = convertHttpUrlToWebsocket(
+        farmPrinters[i].printerURL
+      );
+    }
```
NotExpectedYet, line 1658: Just make sure the webSocketURL key/value actually exists for when people upgrade to this version.

server_src/runners/state.js
```diff
@@ -2028,7 +2030,6 @@ class Runner {
         const { _id } = farmPrinters[index];
         await this.setupWebSocket(_id, skipAPI);
       } else if (
-        farmPrinters[index].ws.instance.readyState === 0 ||
```
NotExpectedYet, line 2031: This was the bugger causing the lock ups when changing the websocket URL.

server_src/runners/state.js
```diff
@@ -3212,9 +3217,9 @@ class Runner {
       });
       let updatePrinter = false;
       if (
-        settings.printer.printerName !== "" &&
-        settings.printer.printerName !==
-          farmPrinters[index].settingsAppearance.name
+        settings.printer.printerName.toString() !== "" &&
+        settings.printer.toString() !==
+          farmPrinters[index].settingsAppearance.name.toString()
```
NotExpectedYet, line 3222: I think these we're why the update was been hooky as I was comparing object -> object.

server_src/runners/state.js
```diff
@@ -3228,14 +3233,28 @@ class Runner {
...
+        currentWebSocketURL.protocol + "//".toString()
+      ) {
+        // If we detect q difference then rebuild the websocket URL and mark for scan.
+        printer.webSocketURL =
+          settings.printer.webSocketProtocol + currentWebSocketURL.host;
+        farmPrinters[index].webSocketURL =
+          settings.printer.webSocketProtocol + currentWebSocketURL.host;
+        printer.markModified("webSocketURL");
+        updatePrinter = true;
+      }
```
NotExpectedYet, line 3257: Update the websocket protocol from the newly selected value

NotExpectedYet (approved):
I've approved as it's your PR but good to check over

server_src/runners/state.js
```diff
@@ -3228,14 +3233,28 @@ class Runner {
       profile.status = 900;
       sett.status = 900;
       if (
-        settings.printer.printerURL !== "" &&
-        settings.printer.printerURL !== farmPrinters[index].printerURL
+        settings.printer.printerURL.toString() !== "" &&
+        settings.printer.printerURL.toString() !==
+          farmPrinters[index].printerURL.toString()
```
davidzwa, line 3238: This hurts my eyes

server_src/runners/state.js
```diff
@@ -3228,14 +3233,28 @@ class Runner {
...
       ) {
         farmPrinters[index].printerURL = settings.printer.printerURL;
         printer.printerURL = settings.printer.printerURL;
         printer.markModified("printerURL");
         updatePrinter = true;
       }
+      const currentWebSocketURL = new URL(farmPrinters[index].webSocketURL);
+      if (
+        settings.printer.webSocketProtocol.toString() !==
+        currentWebSocketURL.protocol + "//".toString()
```
davidzwa, line 3248: "//".toString() ... right

server_src/runners/state.js
NotExpectedYet: Poor planning on sending shit to the backend xD

server_src/models/Printer.js
davidzwa: 😂

server_2x/src/api/legacy_runners/progress/state.js
davidzwa: Those were meant to keep track of the old code and note down what still had to be done. A bad kind of progress tracker tbh, but better than nothing.
Can be removed if its in the way

server_src/runners/state.js
davidzwa: - `settings.printer.printerName` is the string to compare
- you're indeed now comparing an object `settings.printer` to string `farmPrinters[index].settingsAppearance.name` so that is object <-> string.
- please revert the lines where toString was added because it doesnt fix anything. It converts objects to [Object object] or something similar: not what you want at all.

These changes should be cancelled out on dev as this wont work, or does not improve the situation.

server_src/runners/state.js
davidzwa: I dont think you're correct here as well. 

Please choose what applies:
1) `settings.printer.printerURL` is not a string
2) `farmPrinters[index].printerURL` is not a string

If either is the case, then
- what are they?
- why are they that value?
- how does converting a string to string help?

server_src/runners/state.js
davidzwa: This is very weird code

davidzwa:
@NotExpectedYet please take a look at my comments, I think this PR wasnt ready in the last state as I spotted multiple problems which are most likely bugs

server_2x/src/api/legacy_runners/progress/state.js
NotExpectedYet: It's not in the way, just saves updating two sets of files was all I was thinking.

PR #597, opened 2021-05-25
Fix deprecated OP API call in state + backwards compat util

davidzwa:
- [x] 1.6.0 - deprecated plugin API called
- [x] Wrong data taken from call
- [x] Undefined 'settingsAppearance' undefined check bug
- [x] Missing awaits (although async is unecessary)
- [x] Changelog

Furthermore:
- [x] Revert plugin back to repository.plugins
- [x] Fix compat check printer version
- [x] Test compat util

### DIFF
--- a/server_src/runners/state.js
+++ b/server_src/runners/state.js
@@ -2558,7 +2558,7 @@ class Runner {
         await PrinterTicker.addIssue(
           new Date(),
           farmPrinters[index].printerURL,
-          "Grabbed plugin list",
+          `Grabbed plugin list (OctoPrint compatibility: ${farmPrinters[index].octoPrintVersion})`,
           "Complete",
           farmPrinters[index]._id
         );
--- a/server_src/runners/state.js
+++ b/server_src/runners/state.js
@@ -20,6 +20,7 @@ const { JobClean } = require("../lib/dataFunctions/jobClean.js");
 const { FileClean } = require("../lib/dataFunctions/fileClean.js");
 const { FilamentClean } = require("../lib/dataFunctions/filamentClean.js");
 const { PrinterTicker } = require("./printerTicker.js");
+const {checkPluginManagerAPIDeprecation} = require("../utils/compatibility.utils");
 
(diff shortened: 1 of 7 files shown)


server_src/runners/state.js
```diff
@@ -2546,23 +2546,23 @@ class Runner {
     return ClientAPI.getRetry(
       farmPrinters[index].printerURL,
       farmPrinters[index].apikey,
-      "api/plugin/pluginmanager"
+      "plugin/pluginmanager/plugins"
```
NotExpectedYet, line 2549: What about previous versions?

server_src/runners/state.js
```diff
@@ -2546,23 +2546,23 @@ class Runner {
...
     )
       .then((res) => {
         return res.json();
       })
-      .then((res) => {
-        farmPrinters[index].pluginsList = res.repository.plugins;
```
davidzwa, line 2555: @NotExpectedYet did you intend to call the `repository.plugins` here? I'd expect the installed plugins instead of the available ones.

server_src/runners/state.js
davidzwa: I guess we'll have to support both, because the API will become obsolete one day. Shall I do a version check here?

NotExpectedYet: We're going to have to, it would break functionality I gather pre 1.6.0 and we know how fast famers update xD

server_src/runners/state.js
NotExpectedYet: I grab the available ones for the plugin installer.

davidzwa:
image
1.5.3 and 1.5.2

image
1.6.0

NotExpectedYet (approved):
Nice work man

server_src/runners/state.js
```diff
@@ -2745,7 +2745,7 @@ class Runner {
     }
   }
 
-  async static getSettings(id) {
+  static async getSettings(id) {
```
NotExpectedYet, line 2748: xD do this all the time

PR #614, opened 2021-05-30
Bugfix/555 offline after error

NotExpectedYet:
image

This must have been new in OctoPrint 1.6.0 as didn't trigger on 1.5.2

image

This was capturing correctly at least


davidzwa (approved):
It indeed works as you describe. Do note tho that "Error!" gives users very little information as to what is going wrong.
I would personally like to see the message OP sends. The reason is that a refresh results in "Disconnected" instead, so it is quite vague for users as to what is the problem as it is transient only.

NotExpectedYet:
> It indeed works as you describe. Do note tho that "Error!" gives users very little information as to what is going wrong.
> I would personally like to see the message OP sends. The reason is that a refresh results in "Disconnected" instead, so it is quite vague for users as to what is the problem as it is transient only.
> Stuff we can work on in another PR?

There's the log modal you open to see what the actual error is, ain't no way I'm adding the error message to the tiny bubble it's currently in XD. Can't remember if it's on hover but could go there.

davidzwa:
Wasnt thinking of the bubble. Was thinking of adding a custom row below it with small font to display such messages effectively as they are rare but very valuable

NotExpectedYet:
> Wasnt thinking of the bubble. Was thinking of adding a custom row below it with small font to display such messages effectively as they are rare but very valuable. Could also be collapsible
> 
> Printer 1 row
> Small message footer
> Printer 2 row
> Printer 3 row
> etc

Will see about adding it. Could move the API issues / others there.

Issue #615, opened 2021-05-31
Refactor the printer settings saving flow


NotExpectedYet:
### Description of the problem

```bash
Currently there are a lot of patch work here for saving settings between client and server. This can be cleaned up some.
```

### Reproduction of the problem

```bash
...
```

davidzwa:
Note to tackle:
- webcam options patched with settings.other only when `other ` is present

davidzwa:
This is actually fixed now!

PR #616, opened 2021-05-31
Bugfix/fix to string goof

NotExpectedYet:
This PR removes the "toString" conversion and also updates the flow and checks for the printer settings saving. The following now works. 

- Allows the printer settings to update to an incorrect connection profile, page reacts and goes into offline mode. 
- Allows the printer settings to update to a correct connection profile, page reacts and comes back online. 
- Server now correctly checks the status of the printer before continuing with updating. 
- Server will only run the update from OctoPrint when the modal is open if the printer is online. 
- Correctly updates the printer with any changes to connection profiles. 

This PR is not a clean up of the code. That I will save for another PR relating to issue #615


client_src/js/lib/modules/printerSettings.js
```diff
@@ -1402,6 +1402,21 @@ class PrinterSettings {
...
+
+      if (pageElements.menu.printerProfileBtn.classList.contains("notyet")) {
+        pageElements.menu.printerProfileBtn.classList.remove("notyet");
+      }
+      if (pageElements.menu.printerGcodeBtn.classList.contains("notyet")) {
+        pageElements.menu.printerGcodeBtn.classList.remove("notyet");
+      }
+      if (pageElements.menu.printerOtherSettings.classList.contains("notyet")) {
+        pageElements.menu.printerOtherSettings.classList.remove("notyet");
+      }
```
NotExpectedYet, line 1419: Now allows the page to react to printer going Offline -> Online

server_src/runners/state.js
```diff
@@ -3272,13 +3274,18 @@ class Runner {
...
-        settings.connection.preferredBaud;
-      farmPrinters[index].options.portPreference =
-        settings.connection.preferredPort;
-      farmPrinters[index].options.printerProfilePreference =
-        settings.connection.preferredProfile;
+      farmPrinters[index].options = {
+        baudratePreference: settings.connection.preferredBaud,
+        portPreference: settings.connection.preferredPort,
+        printerProfilePreference: settings.connection.preferredProfile
+      };
```
NotExpectedYet, line 3288: This needs blank checking as well currently. Don't see the point at the moment, have added a note for doing in the refactor as the rest needs to be changed too. I'll make the client only send updated values, so the server only has to react to available values.

server_src/runners/state.js
```diff
@@ -3409,29 +3416,21 @@ class Runner {
...
-      } else {
-        settings.systemCommands.systemRestart =
-          farmPrinters[index].settingsServer.commands.systemRestartCommand;
       }
       if (settings.systemCommands.systemShutdown !== "") {
         farmPrinters[index].settingsServer.commands.systemShutdownCommand =
           settings.systemCommands.systemShutdown;
-      } else {
-        settings.systemCommands.systemShutdown =
-          farmPrinters[index].settingsServer.commands.systemShutdownCommand;
```
NotExpectedYet, line 3428: legacy code from an even older update method, these we're all pointless as I'm re-updating what was sent.

server_src/runners/state.js
```diff
@@ -3409,29 +3416,21 @@ class Runner {
...
       }
 
       printer.save().catch((e) => {
         logger.error(JSON.stringify(e), "ERROR savin power settings.");
       });
-      if (settings.state !== "Offline") {
+      // Made the state check from the server, not the client...
+      if (farmPrinters[index].state !== "Offline") {
```
NotExpectedYet, line 3433: This was causing the issue of the client not rescanning after a URL change. The client was sending "Offline" state so it was never triggering.

server_src/runners/state.js
```diff
@@ -3459,12 +3458,13 @@ class Runner {
...
-            webcamEnabled: settings.other.enableCamera,
-            timelapseEnabled: settings.other.enableTimeLapse,
-            rotate90: settings.other.rotateCamera,
-            flipH: settings.other.flipHCamera,
-            flipV: settings.other.flipVCamera
+            webcamEnabled: settings.other?.enableCamera,
+            timelapseEnabled: settings.other?.enableTimeLapse,
+            rotate90: settings.other?.rotateCamera,
+            flipH: settings.other?.flipHCamera,
+            flipV: settings.other?.flipVCamera
```
NotExpectedYet, line 3467: These are here purely because currently when you send a save command with an offline printer the "other" object is never sent. Will clean up in the refactor.

NotExpectedYet:
@davidzwa  have annotated.

server_src/runners/state.js
davidzwa: Sounds better to leave out the webcam section if `settings.other` is missing. The only reason this now doesnt change webcam is because `undefined` properties are removed in JS

NotExpectedYet: Yeah I agree, I will tackle that in the cleanup PR as currently what the client sends it a mess.

davidzwa:
**Problem**
After saving printer connection settings like f.e. port (doesnt matter if changed or not) the save action results in the printer showing in the modal as Disconnected and Websocket tentative.
image

**Repro**
Saving printer connection settings in the modal.
When closing the modal, the printers table does show OP "Connected" and websocket connected.
When re-opening the dialog the state is initially still Disconnected and WS tentative for about a sec or 2.
After that the modal syncs and the state changes to what is correct.

**Expected behaviour**
When saving the settings I expect the connection state to be updated to the last state eventually.

davidzwa (changes requested):
See bug report above

NotExpectedYet:
Hmmm, it's refreshing the wrong printer

NotExpectedYet:
Fixed the bug @davidzwa 

The page refresh/save function I was using was overriding the printerInformation array from the websockets which was causing the printer Index to get mismatched. On the next tick from the SSE the full printerInformation array comes through and it uses the 0 index due to that goof. 

The first tick will now update the currentPrintersInformation for usage in those functions. This is required as the SSE event would have the latest information that the functions need to know about. Passing through wouldn't have that, it would contain the printersInformation array that was pass through on opening the modal. 

I also made the page refresh happen on the initial tick rather than awaiting for the second tick of SSE so should update sooner now and not show the old information.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ All notable changes to this project will be documented in this file.
     - Made API tests less prone to unnecessary failure
     - Reworked the Settings modal to be more resiliant to failure and cleaned up code
     - Slightly reworked job cache, prepared for V2
+    - Added the ability to override the automatic wss:// select for HTTPS in printer settings modal. 
+    - Added the ability for settings dialog to return to "Printer Online" view when printer comes online in background / from settings changes.
     - Amended the functions for Global OctoPrint update and Global OctoPrint plugin update
     - Added the ability to override the automatic wss:// select for HTTPS in printer settings modal
     - Added the ability to override the automatic wss:// select for HTTPS in printer settings modal
```
davidzwa, line 21: Duplicate

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ All notable changes to this project will be documented in this file.
     - Made API tests less prone to unnecessary failure
     - Reworked the Settings modal to be more resiliant to failure and cleaned up code
     - Slightly reworked job cache, prepared for V2
+    - Added the ability to override the automatic wss:// select for HTTPS in printer settings modal.
```
davidzwa, line 17: Duplicate

NotExpectedYet: You duplicated your duplicate comment

server_src/lib/dataFunctions/printerClean.js
```diff
@@ -416,7 +416,7 @@ class PrinterClean {
       );
       sortedPrinter.connectionOptions = farmPrinter.options;
       if (
-        !!sortedPrinter?.connectionOptions &&
+        !!sortedPrinter?.connectionOptions?.ports &&
```
davidzwa, line 419: Bugfix. For my VIRTUAL only OP instance there was no ports property. Could be that this array is not sent when only 1 connection exists? OP version 1.6+

server_src/services/octoprint/octoprint-api.service.js
```diff
@@ -125,7 +125,8 @@ class OctoprintApiService {
       url,
       "PATCH",
       apiKey,
-      timeout ? this.timeout.apiTimeout : false
+      timeout ? this.timeout.apiTimeout : false,
+      data
```
davidzwa, line 129: My doing

davidzwa:
@NotExpectedYet I had to fix a bug in `statejs` I created. We're calling PATCH for the printer profile, but no data was sent as body. That was my doing, which is now fixed here.
My question to you is why the data we're sending to update the OP printer profile does not match the OP REST API?
https://docs.octoprint.org/en/master/api/printerprofiles.html#update-an-existing-printer-profile

The code where the profile is patched, is found here:
https://github.com/OctoFarm/OctoFarm/blob/a11c54f7f2f9aca8fb7fea54b2af7ce106f84262/server_src/runners/state.js#L3371

davidzwa (approved):
LGTM, can advise to check my changes as well.

NotExpectedYet:
Yeah will double check before merge.

PR #631, opened 2021-06-18
Speed up Windows performance x5-x10

davidzwa:
- [x] Improve situation by disabling SystemRunner completely
- [x] Add back functionality for system page in optional dialog (no Runner!!)
- [x] Fix tests (probably)
- [x] Reduce and make compact system info usage
- [x] Changelog
- [x] Compensate for concessions (CPU, memory)


davidzwa:
WSL Linux proof
image

I can also prove (but to lazy to GIF it) that the chart updating works on Windows. It never did on WSL/docker so nothing out of the ordinary there.

Windows before query
image
After
image
After2 (proof that CPU changed)
image

NotExpectedYet:
ShareX wicked gif creator :P

NotExpectedYet (approved):
Tested locally and checked through all looking good. 

Initial findings... 

Everything loads much faster, even on linux. Calls on every page are resolved quicker. Nice find dude!

PR #633, opened 2021-06-19
🐛 regression printer settings module exports

davidzwa:
Not gonna changelog as this is a regression by #545 in the same release cycle. I think this isnt triggered because gulp with babel skips it? Not sure, but the webpack PR failed.

Going to add a task for adding frontend tests later. This will help us migrate to V2 client


client_src/js/lib/modules/printerSettings.js
```diff
@@ -1424,6 +1428,3 @@ class PrinterSettings {
     }
   }
 }
-module.exports = {
-  updatePrinterSettingsModal
-};
```
NotExpectedYet, line 1429: I use this export in printer manager

davidzwa: URL?

NotExpectedYet: https://github.com/OctoFarm/OctoFarm/blob/55a56934613d9fca1208a4743a082f97e926a899/client_src/js/printerManagementRunner.js#L6

NotExpectedYet (approved):

Issue #638, opened 2021-06-19
Login POST route kaputt on dev - fix passport regression


davidzwa:
### Description of the problem

```bash
Self-explanatory. Add more true login tests to prevent this in future.
```

Caused by #601 we didnt test properly with login required set to false and a true logout.

Issue #589, opened 2021-05-19
Persist login token across restarts


davidzwa:
### Description of the problem

```bash
Save the login token across restarts of the OF server.
```

### Reproduction of the problem

```bash
Login is reset and old session cookies are bounced after OF restart.
```

PR #477, opened 2021-04-25
Bugfix/settings improvement cache

NotExpectedYet:
Think I'm gonna do this in a single PR and get it into 1.1.13 as it's hugely beneficial regarding the missing settings. 

Future plans are to re-work the rest of the database and move settings into a better cache set up outside of lib/ and into services. 

Tested by nuking a load of default settings and keys in MongoCompass and restarted the server and they we're back to the default settings :D.

PR #639, opened 2021-06-19
🐛 #589 fix login - investigate persistence (skipped persistence ... for reasons)

davidzwa:
This PR fixes the login regression #638 and adds tests to monitor it during coding.

I tried this as well:
- Introduce cookieSession instead of session
- Main test fails
- Postman fails
- Browser works
- Go figure

So... persisting is easy, but testing is not.... should we ignore tests and just go for `cookieSession`?


NotExpectedYet (approved):
- [x] No login required, all pages load. Root redirects to dashboard...
- [x] Login required, logs in and all pages load. Root redirects to dashboard... 
- [x] Logged out, pages redirect to welcome
- [x] Can register user
- [x] Remember me works
- [x] First user created is administrator
- [x] Correct error messages for login form. Passwords don't match, missing fields.

NotExpectedYet:
> This PR fixes the login regression #638 and adds tests to monitor it during coding.
> 
> I tried this as well:
> 
>     * Introduce cookieSession instead of session
> 
>     * Main test fails
> 
>     * Postman fails
> 
>     * Browser works
> 
>     * Go figure
> 
> 
> So... persisting is easy, but testing is not.... should we ignore tests and just go for `cookieSession`?

From what you tested and what I tested you only really missed the register form no? 

I listed what I tested this time for you at least. Could always bolster the tests more.

davidzwa:
The registration form is used twice: page load and POST. Its not fully tested w.r.t. server settings, but thats up to #477 to cover. If you could do that it would be great

Oh and the `cookieSession()` middleware works fine across server restarts instead of `session`. The thing is that tests start to fail as somehow only the browser is accepted. We should look into fixing the browser vs postman/tests and switch.

PR #654, opened 2021-06-26
Bugfix/fix noty undefined

NotExpectedYet:


davidzwa (approved):

davidzwa:
Like this solution better, thanks

PR #655, opened 2021-06-26
Bugfix/sse flatted new major version namespace

davidzwa:
The `flatted/cjs` endpoint had to be changed to `flatted` for version 3 of this package.


NotExpectedYet:
Worth updating the client too? It uses my parsed file as I could never import it from node_modules before.

davidzwa:
Ah righto, yeah sounds good

davidzwa:
@NotExpectedYet shall we merge this? The client side I dont have the time for.
Repercussion: SSE will fail when V2 flatted is installed instead of V3. Write a fallback?
Alternative: revert

NotExpectedYet:
> @NotExpectedYet shall we merge this? The client side I dont have the time for.
> Repercussion: SSE will fail when V2 flatted is installed instead of V3. Write a fallback?
> Alternative: revert

sorted client, just waiting for checks to run

NotExpectedYet (approved):

PR #662, opened 2021-06-30
🚀 Chore/debug gha profiles find

davidzwa:
- [x] Test run the GHA to fix profiles.find bug and test startup issues
- [x] Fix tests after test breaking for dashboard SSE
- [x] Discuss with The Brit how to proceed with startup mumbo jumbo
- [x] Change some logs


app-core.js
```diff
@@ -155,7 +154,36 @@ async function serveOctoFarmNormally(app, quick_boot = false) {
...
+    await PrinterClean.sortCurrentOperations(printersInformation);
+    await PrinterClean.statisticsStart();
+    await PrinterClean.createPrinterList(
+      printersInformation,
+      serverSettings.filamentManager
+    );
+
+    // let interval = false;
+    // if (interval === false) {
+    //   interval = setInterval(async () => {
```
davidzwa, line 177: @NotExpectedYet I've not yet introduced a periodic call here. Do we have anything which is more abstract than setInterval?
Is this robust against errors f.e.? And what about suspending/stopping the asynchronous code or managing this during tests? Got any idea for that? (Im not mentioning a background job runner to not immediately force us in that train of thought)

NotExpectedYet: Not that I'm aware of. The only way to repeat a function that I'm aware of would be with setInterval or a function call loop and a setTimeout to trigger the next call of the function. A nested setTimeout would be more controllable than a setInterval. 

We could write something that does those functions but I wonder if a library exists.

davidzwa: Im thinking about writing a manager to manage the intervals.
1) Request a time interval + pass a sync/async callback
2) Manager saves/memorizes the interval number and can restart it/kill it/idle it

davidzwa:
@NotExpectedYet I can confidently say that the tests are 100% stable now with this task manager in place. I still need to uncomment the `await FilamentClean.start();` line as that is possibly one of the culprits.

Might have to dig deeper to test. What do you think by looking at the changes?

NotExpectedYet:
> @NotExpectedYet I can confidently say that the tests are 100% stable now with this task manager in place. I still need to uncomment the `await FilamentClean.start();` line as that is possibly one of the culprits.
> 
> Might have to dig deeper to test. What do you think by looking at the changes?

I'm liking the changes so far yeah. The task scheduler is a great additon! 

Filament clean to me (thinking about it) is something that can probably just be on demand and skipped from start up all together.  I think it is already on demand, I'd have to double check to see if that's actually the case.

NotExpectedYet (approved):

PR #666, opened 2021-07-02
Bugfix/prerelease gulp problem

davidzwa:
Adjust frontend build to push at a tag/push to master/dev and only dry-run for PR.
Remove gulp prerelease command


NotExpectedYet (approved):
Smash it

PR #671, opened 2021-07-02
Bugfix/688-670 api fix

NotExpectedYet:
Defaulted these back to true as they we're before the refactor. 

These can be utilised further in the future once the state.js actually sends the flags. Currently it's not sent at all for files or connection attempts and some of these were fixes for OP clients on Wifi that sometimes didn't respond within 1000ms. 

The files one only sends the recursive false flag on a single file update/scan currently, this also needs to be true. Probably should have noticed this when I was reviewing. XD

### DIFF
--- a/server_src/routes/printers.js
+++ b/server_src/routes/printers.js
@@ -73,7 +73,7 @@ router.post("/resyncFile", ensureAuthenticated, async (req, res) => {
   if (typeof file.fullPath !== "undefined") {
     ret = await Runner.reSyncFile(file.i, file.fullPath);
   } else {
-    ret = await Runner.getFiles(file.i, "files?recursive=true");
+    ret = await Runner.getFiles(file.i, true);
   }
   setTimeout(function () {
     res.send(ret);
--- a/server_src/runners/state.js
+++ b/server_src/runners/state.js
@@ -2141,7 +2141,7 @@ class Runner {
     const printer = farmPrinters[index];
 
     return await this.octoPrintService
-      .getFiles(printer, true, true)
+      .getFiles(printer, recursive)
       .then((res) => {
         return res.json();
       })
--- a/server_src/runners/state.js
+++ b/server_src/runners/state.js
@@ -2675,8 +2675,7 @@ class Runner {
           );
 
           let piSupport = await this.octoPrintService.getPluginPiSupport(
-            farmPrinters[index],
-            true
+            farmPrinters[index]
           );
           piSupport = await piSupport.json();
           logger.info("Got from endpoint: ", piSupport);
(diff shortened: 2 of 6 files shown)


NotExpectedYet:
Hmmm the tests did not like that

NotExpectedYet:
I believe this was due to the api key check not requesting a timeout and not increasing to the secondary timeout

Made the following changes anyway. Would like confirmation from the user that this resolved the issue. 

- [x] Updated state.js file manager calls to send true/false boolean for recursive rather than the string.
- [x] Tweaked the ticker to show human readable times (seconds not ms)
- [x] Made the initial timeout be overridden by the secondary timeout, this will reduce subsequent calls down to 1 call rather than 2 consistently as before. If the first timeout fails then it's always going to fail at 1000ms. 

So the API endpoints will now make a call, if the first of these calls fails at 1000ms it will try again straight away at 10000ms. These are both user controllable from the System page. Any subsequent call will use the new values of 10000ms (apiRetry) setting. Seems I made a goof some time ago as apiRetry and apiRetryCutOff are switched. Left for now.

 - Call, timeout of 1000ms reached, retry + increase to secondary timeout value.
 - Call again, timeout reached, hits cut off value and errors out.
 - Printer is considered offline and doesn't re-attempt for another 30 seconds. 
 - 30 seconds pass, Call again but with new timeout value of 10000ms. 

With changes:
image
Successfully retries and gives up when 10000ms timeout exceeded. 

Without changes:
image
Fails to retry correctly at the api key check and never attempts another connection until 30 seconds has passed.

server_src/routes/printers.js
```diff
@@ -73,7 +73,7 @@ router.post("/resyncFile", ensureAuthenticated, async (req, res) => {
   if (typeof file.fullPath !== "undefined") {
     ret = await Runner.reSyncFile(file.i, file.fullPath);
   } else {
-    ret = await Runner.getFiles(file.i, "files?recursive=true");
+    ret = await Runner.getFiles(file.i, true);
```
davidzwa, line 76: Wow really bad. Good find!!

server_src/runners/state.js
```diff
@@ -2675,7 +2675,8 @@ class Runner {
           );
 
           let piSupport = await this.octoPrintService.getPluginPiSupport(
-            farmPrinters[index]
+            farmPrinters[index],
+            true
```
davidzwa, line 2679: Did a quick test for my WLAN Prusa OP. This API endpoint responds in 50 to 100ms. So I wonder why is a retry is needed?

Look I completely understand why we need the retry mechanism somewhere, but to be frank I'd only see its use for a OP connection test to keep the connection/power state of that OP service updated. Other than that I see no point in retrying anywhere. (Didnt read your PR analysis yet, quite tired so will try to do so now haha)

davidzwa (approved):

PR #673, opened 2021-07-04
🐛 Bugfix/590 669 file manager fixes

davidzwa:
- DISREGARDED as too complex ~Sort dropdown only expands on arrow (not on clicking text)~
- [x] Back button doesnt gray out for root folder
- [x] Add left-right border color to selected printer
- [x] Printer name made smaller by adding primary badge
- [x] Printer name in title made consistent by adding primary badge
- [x] Print cost rounded toFixed(2) client side with `?` operator
- [x] File Body height was too large for empty list (kept minimum height of 40vh as easy fix, thats accounts for 2 printers in the list)
- [x] No files now will result in "No files to be shown"
- [x] No files now will result in "No files to be shown" for subfolder
- [x] Filter file list based on current Folder before, not after constructing the element to be inserted
- [x] Scroll bar is used when not needed
- [x] Scroll isnt reset when switching printers
- [x] Changelog 

Client bundle is pushed and versioned `0.1.0-rc2-build1`

Post review fixes:
- [x] Secondary badge colors (2x)
- [x] Var name corrected
- [x] Changelog more according to truth


davidzwa:
> * DISREGARDED as too complex ~Sort dropdown only expands on arrow (not on clicking text)~
image
Wanted to merge these as it annoyed me, but decided not to.
> * [x]  Back button doesnt gray out for root folder
image
Graying out logic applied for root folder Local and subfolders (true/false logic)
> * [x]  Add left-right border color to selected printer
See below
> * [x]  Printer name made smaller by adding primary badge
image
As promised
> * [x]  Printer name in title made consistent by adding primary badge
image
> * [x]  Print cost rounded toFixed(2) client side with `?` operator
image
> * [x]  No files now will result in "No files to be shown"
See below
> * [x]  No files now will result in "No files to be shown" for subfolder
Tested. Works fine. See below.
> * [x]  File Body height was too large for empty list (kept minimum height of 40vh as easy fix, thats accounts for 2 printers in the list)
image
> * [x]  Filter file list based on current Folder before, not after constructing the element to be inserted
> * [x]  Scroll bar is used when not needed
Fixed see above
> * [x]  Scroll isnt reset when switching printers
Fixed

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
     - Fixed #665: If Global API check fails due to intial time out, never recovered and tried again increasing timeout.
     - Fixed #670: File manager initial and subsequent scans we're not recursive. 
+    - Fixed #590 #669: File manager scroll wouldnt reset after switching printer. Nice visual tweaks and the Back button logic.
```
NotExpectedYet, line 17: Could you add something about the view changes as well please? 

I'll add file manager to one of the "Cleanups" too :)

client_src/js/fileManagerRunner.js
```diff
@@ -56,7 +56,7 @@ class Manager {
               <div class="col-lg-10">
                 <div class="d-flex w-100 justify-content-between">
                   <h5 class="mb-1">
-                    <span id="printerName-${printer._id}">
+                    <span class="badge badge-primary" id="printerName-${printer._id}">
```
NotExpectedYet, line 59: Can this be secondary rather than primary, to fit with the other title's in the UI? 

Blue sticks out like a saw thumb in my opinion.

client_src/js/fileManagerRunner.js
```diff
@@ -174,8 +173,13 @@ class Manager {
 
   static changePrinter(e, target) {
     if (!e.target.id.includes("filamentDrop")) {
+      const filamentList = document.getElementById("fileBody");
+      if (!!filamentList) {
+        filamentList.innerHTML = "";
+        filamentList.scrollTop = 0;
+      }
+
```
NotExpectedYet, line 181: I can gather this is my earlier goof but filamentList?

NotExpectedYet: What was I smoking XD

NotExpectedYet (changes requested):
LGTM just left some comments/changes. 

Haven't tested though and don't know when I'll get chance at the moment. 

Based on the changes been made and it still keeping functionality in "File Manager" and "Printer Control" then I will approve.

client_src/js/fileManagerRunner.js
NotExpectedYet: Oh same for the title as well

client_src/js/fileManagerRunner.js
davidzwa: Oooh this is all me, except for the line before where "filamentDrop" is shown. You know those situations where contextual words make you type the wrong thing? This is my goof hahaha. Will fix later, as well as your other comments.

NotExpectedYet: Omg, not me! ahaha no worries

client_src/js/fileManagerRunner.js
davidzwa: I find the secondary color quite weak in all the grays already present in the printer list. Secondly, there's little to nothing visually linking the selected printer to the shown file list except similar text (which isnt the same size and not the same transparency).
That's why I made it blue, for me it doesnt stand out as the blue is quite dim.

Try to look at the text I marked and see the problem with `badge-secondary`
image

davidzwa: Oh and another point: the list hover color is not consistent making it a bit uncomfortable for me. Plus the fact that the cursor constantly changes, Im a bit on the edge of my seat there

NotExpectedYet: I agree the background now looks out of place on selected. That's originally why I had it as a text field with transparent background and not a badge/button. Changing that style seems a bit much now I've seen it in the screenshot, Maybe just revert to making both texts the same as the others on the file navigation bar? What do you think? 

I still think the "Files: # / Folders: #" text would go nicely under the "Operational" badge, did you try that at all? So would be 
"Operational"
"Files: #"
Folder #" 
and would utilise that space there a bit more rather than just having the printer icon. 

Personally regarding the colours I'd not colour the texts background as it's a stark difference to every other card around the interface. Maybe change the selection colour of the card? They alternate between grey and darker grey so a colour that stands out more may be better suited there. 

I've never had any complaints over readability there so maybe I'm just a bit blind sided by the "This is obviously the selected printers card list" so feel free to whack my out of that. Screenshots certainly help aha.

NotExpectedYet: Regarding the loading I've never personally noticed the long load times, even with searching which pulls the complete file list before filtering (this would still be needed as the search is recursive, so maybe a server side function for that). I wouldn't personally say pagination is required but having it just load the current directory rather than the whole structure is needed if there is indeed slow loading times. Would you say that's relevant to this PR? I'm leaning against not but 100% something that needs to be added.

NotExpectedYet: History needs the same as well

davidzwa: About pagination: I dont mind deferring that. Would be nice to be able to benchmark such calls with the runOnce functionality I had in mind on the other `profiles.find`/task scheduler PR.

About the files/folders, I didnt realize you opted for the left: I like it. Will apply on this PR later.

About the printer selection, I would make the selected printer less dark and make the hover the same or nearly the same color (now its quite a difference).

About the current printer, there's also another option. We can make the other text on the file list title bar a different text to show those are "details" but the printer name is not/important. F.e. using <small> on the storage details or <strong> on the printer name. I do hope you agree we need something more than the same font and size. The problem is printer names can be long which means a big text shows up on small space.

About graycolors, I'd look for flattening the grays used a bit. Look at github dark mode for what I envision. It has a blue navbar and dark background. The rest is borders and some panel background gray. In total about 3/4 grays max. OF has about 6. I'd propose to bring that back to 3/4.

Sidenote: changing the selected printer or clicking a subfolder shifts the page. Hope to fix that. Annoys me a bit.

NotExpectedYet: > About pagination: I dont mind deferring that. Would be nice to be able to benchmark such calls with the runOnce functionality I had in mind on the other `profiles.find`/task scheduler PR.
> 
> About the files/folders, I didnt realize you opted for the left: I like it. Will apply on this PR later.
> 
> About the printer selection, I would make the selected printer less dark and make the hover the same or nearly the same color (now its quite a difference).
> 
> About the current printer, there's also another option. We can make the other text on the file list title bar a different text to show those are "details" but the printer name is not/important. F.e. using on the storage details or ** on the printer name. I do hope you agree we need something more than the same font and size. The problem is printer names can be long which means a big text shows up on small space.**
> 
> ** About graycolors, I'd look for flattening the grays used a bit. Look at github dark mode for what I envision. It has a blue navbar and dark background. The rest is borders and some panel background gray. In total about 3/4 grays max. OF has about 6. I'd propose to bring that back to 3/4. Sidenote: changing the selected printer or clicking a subfolder shifts the page. Hope to fix that. Annoys me a bit.**

Was on discord, no drama if you missed it regarding the left. 

I'm happy to see what you come up with regarding what you said. I'm in agreement it needs something more totally. 

Yeah the printer names is something I've struggled with quite a bit, I personally think truncating them anything more than an IP+Port length would suffice. Just going off mostly what I've seen of my users they like to call their printers cartoon names or something similar, but I'd gather larger farms would more go for a server naming convention so I've always been reserved to cut off the last set of letters as that's usually where the identifying part is.

davidzwa: What about setting an avatar for each printer and/or using the first letter + a random color as a identifier if no avatar is set?
The idea is here, taken from Azure DevOps (or microsoft single sign on)
image

Discord does this brilliantly:
image

I guess we have much more creative room than you might think!

NotExpectedYet: My only issue with that is it's build around names which are generally two strings added togather so you get two sets of first letters. A farm using location codes wouldn't fit very well into that as they'd probably have rack-1-2 or colour designations. I also don't think it covers the character naming side as just for example, Micky/Mini would look the same. One more thing! IP's wouldn't work as you'd get "H" or "1" normally. 

I do agree there's more creative room here than is currently displayed in my early attempts. xD

davidzwa: The thing is if we offer a way to organize, users will most probably use it and sanitize their printer names to fit. We just have to choose and take stacking/grouping/location placement into account.

Oh and Minnie/Miny => Mi#1 Mi#2

NotExpectedYet: > The thing is if we offer a way to organize, users will most probably use it and sanitize their printer names to fit. We just have to choose and take stacking/grouping/location placement into account.
> 
> Oh and Minnie/Miny => Mi#1 Mi#2

As someone who like's their custom naming scheme I would be annoyed at Mi#1 Mi#2 baha. Is Mini 1 or Micky 2, requires me to remember an order. It's something we need to discuss going forward anyway, and a user choice for what's displayed is always a good idea. It's worth a discussion maybe? then an issue when it's been worked out as it's a bit disjointed around the views currently.

davidzwa: https://github.com/OctoFarm/OctoFarm/discussions/677

NotExpectedYet:
I've pushed some tweaks to this. 

1. these not matching was annoying my OCD. Should be level in my opinion. 
image
image

2. Gave the text a transparent background with a disabled button like the rest of the UI. This also wraps the text as per my screenshot with bob. 
image

3. Move file and folders here: 
image

4. Made these back the same size
image

5. Added some spacing under extruders
image

client_src/css/octofarm.css
```diff
@@ -280,14 +280,16 @@ a.notyet {
...
   max-height: 65vh;
 }
-
 .fileBody p {
   margin-bottom: 0px !important;
 }
-
+.printerSelected {
+  border-left: #f39c12 4px solid;
+}
```
NotExpectedYet, line 288: Do like how this looks

NotExpectedYet (approved):
Love this dude, nice one. 

If you approve my little tweaks then it's approved.

PR #680, opened 2021-07-11
Bugfix/679 disable op update check airgapped

NotExpectedYet:
draft until 678 is merged

- [x] return false on plugin update and octoprint update checkers
- [x] disable the plugin installation button when airgapped. 

Tested by forcing true on air_gapped variable and worked fine. Didn't take 10 seconds for a printer to scan.


davidzwa (approved):
lgtm with the latest changes

We discussed the synchrony of air_gapped and added #682 to solve it for good.

PR #685, opened 2021-07-11
Bugfix/605 null value null when no spool

NotExpectedYet:
Fixed! xD


davidzwa (approved):
First change is prolly my doing? I think?
Second... probably as well......... whoops

PR #686, opened 2021-07-11
Fixed #667 Weekly Utility was not saving previous days values.

### LINKED ISSUE #667: Weekly Util is broken
### Description of the problem

Isn't reading historical and isn't saving to database by the looks

### Reproduction of the problem

```bash
...
```

NotExpectedYet:


NotExpectedYet:
brokey tests whoops

NotExpectedYet:
@davidzwa I've commented out the tests for now as wasn't exactly sure why they we're failing

davidzwa:
> @davidzwa I've commented out the tests for now as wasn't exactly sure why they we're failing

I fixed the tests and it shows that running the heatmap function a second time still results in NaN.
The first call returns 0 now, which is nice.

NotExpectedYet:
@davidzwa more work to do then, will take a look

davidzwa (approved):
Tests are not running in predictable fashion: running the previously failing test on its own fails, but running it in sequence results in different heatmap state... funky.

I'd hope it would be 100% predictable. Up to you what you can do with it.
NaN still present in figure default value. Up to you to decide whether relevant.

I get Wednesday and Sunday, I'd expect to see the whole week... any idea?

NotExpectedYet:
@NotExpectedYet I think it's time I moved it into task manager and redid the flow. 

You should get whatever day your on. 

I think all I fixed here was the not using the database values so it was always at a single day. 

Looking at my rc1 version that's ticking along nicely though. No better time to clean it up in my opinion.

PR #688, opened 2021-07-13
Feature/task manager improve

davidzwa:
As promised, with user-friendly changelog addition.

- Cleanup logging with the idea that only crashes or truely valuable information is shown
- Flatten task manager/scheduler combo
- Add run-once, run-delayed + configuration and validation options

Check out `server_src/task.presets.js` and `server_src/tasks.js`.


davidzwa:
@NotExpectedYet lets get this merged so you can work with the task manager.
The promise in the changelog about bringing back logging I intend to keep, just one thing at a time. I think we both agree we can log more condensed information or selectively.

F.e. a logging configuration file just like OP could work.

NotExpectedYet (approved):

PR #711, opened 2021-07-25
Chore/axios global error handler

NotExpectedYet:
This creates an axios error handler and also lays ground work for more axios based validation. 

@davidzwa be happy to get some pointers on this. 

Still to do: 
 - [x] catch known api error codes
 - [x] update client to use new error codes

image


davidzwa:
lgtm

PR #714, opened 2021-07-25
Missing throws

davidzwa:
Noticed these before, didnt submit the review.
On the client changes - its quite hard to adjust the base variable and sometimes its required. So I usually set it in a common variable just like you did, but then use it in another route variable.

```
base = "/api";
printerRoute = urlMerge(base, "/printer"); // some slash safe merging tool
settingsRoute = urlMerge(base, "/settings"); // some slash safe merging tool
weirdRoute = "/weirdBase/weirdPath"; // Still an option without having to change get/post/put/patch/delete.
```


NotExpectedYet (approved):

NotExpectedYet:
> Noticed these before, didnt submit the review.
> On the client changes - its quite hard to adjust the base variable and sometimes its required. So I usually set it in a common variable just like you did, but then use it in another route variable.
> 
> ```
> base = "/api";
> printerRoute = urlMerge(base, "/printer"); // some slash safe merging tool
> settingsRoute = urlMerge(base, "/settings"); // some slash safe merging tool
> weirdRoute = "/weirdBase/weirdPath"; // Still an option without having to change get/post/put/patch/delete.
> ```

Yeah need to sort those out on that settings PR will sort.

client_src/js/services/octofarm-client.service.js
```diff
@@ -99,27 +98,27 @@ export default class OctoFarmClient {
   }
 
   static async get(path) {
-    const url = new URL(this.base + path, window.location.origin).pathname;
+    const url = new URL(path, window.location.origin).href;
```
NotExpectedYet, line 101: These are better as pathname to support reverse proxy instances...

davidzwa: Okido will revert

NotExpectedYet: I've just double checked the documentation and pathname won't work if we ever want to send `?=` queries so can be left as href as I think I was wrong with my reasoning before. Javascript should just detect what's in the url bar regardless.

davidzwa: Yeah note I removed the params static. Got a bit of a different implementation in mind where each function like getPrinter(id)
constructs its own url from start to finish `${base}/${path}?param1=asd`
Just thought to let you know.

PR #720, opened 2021-08-02
Lots of small fixes and constants added

davidzwa:
- V2 Remove NestJS legacy assets
- constants moved or expanded
- added the awilix container and the di tokens (unused!)
- client nullability checks
- moved some utils + tests

Client pages load fine.


client_src/js/lib/octoprint.js
```diff
@@ -13,7 +13,8 @@ export default class OctoPrintClient {
 
   static get(printer, item) {
     this.validatePrinter(printer);
-    const url = `${printer.printerURL}/${item}`;
+
+    const url = new URL(item, printer.printerURL).href;
```
davidzwa, line 17: @NotExpectedYet let me know if this conflicts. If so Ill revert it from this PR.

NotExpectedYet: @davidzwa It will do as I removed all the get's out of octoprint-client.js. I also changed it's name and put it in services. 

The URL checking is done before axios is called as well so you don't need to repeat it. Was going to try and do the same for validate printer too.

client_src/js/services/octofarm-client.service.js
```diff
@@ -65,6 +66,19 @@ export default class OctoFarmClient {
...
+  static validateRequiredProps(input, keys) {
+    const unsetRequiredProps = keys.filter((prop) => {
+      return !input[prop];
+    });
+    if (unsetRequiredProps.length) {
+      console.error(
+        "The following properties were empty/missing in the request",
+        unsetRequiredProps
+      );
+      throw new ApplicationError(ClientErrors.FAILED_VALIDATION_PATH);
```
NotExpectedYet, line 78: key or path here? aha or just re-using my goof?

server_src/constants/state.constants.js
```diff
@@ -130,38 +95,155 @@ const OP_STATE = {
...
 
+const OF_STATE_REMAP = {
+  [OP_STATE.Offline]: {
+    state: PSTATE.Disconnected, // hard remap!
+    desc: "Your printer is disconnected"
+  },
+  [OP_STATE.OpeningSerial]: {
+    state: PSTATE.Searching, // Lack of better
+    desc: "Your printer is connecting to serial"
+  },
```
NotExpectedYet, line 141: Just curious why these are wrapped in an array?

NotExpectedYet: Seen down below.

davidzwa: This is how you can define an object key when using a variable. Glad you asked! Found this trick half a year ago when doing the same kind of work for the rugby app

client_src/js/services/octofarm-client.service.js
davidzwa: Re-using. I was utterly confused xD

server_src/constants/state.constants.js
```diff
@@ -130,38 +95,155 @@ const OP_STATE = {
...
   }
+  if (state === PSTATE.Online) {
+    return { name: "success", hex: "#00330e", category: CATEGORY.Idle };
+  }
   if (state === PSTATE.Paused) {
     return { name: "warning", hex: "#583c0e", category: CATEGORY.Idle };
   }
   if (state === PSTATE.Printing) {
-    return { name: "warning", hex: "#583c0e", category: CATEGORY.Active };
+    return { name: "success", hex: "#583c0e", category: CATEGORY.Active };
```
NotExpectedYet, line 246: This should be "warning" sucess for printing states is only used for a print completed. Think it's confusing to the user those both been the same.

server_src/constants/template.constants.js
```diff
@@ -0,0 +1,5 @@
+const noSpoolOptionTemplate = `<option value="0">No Spool Selected</option>`;
+
+module.exports = {
+  noSpoolOptionTemplate
+};
```
NotExpectedYet, line 5: note to self, shift this stuff client side

davidzwa: Nice plan of attack: filter/split it out and move client side in one go?

server_src/constants/state.constants.js
davidzwa: I do agree, wasnt supposed to be in this PR. Good catch. I do think we should not be limited to the limited set of colors by bootstrap. (Also given the Contrast Ratio test failing). 

Will revert

NotExpectedYet (changes requested):
Just some questions and 1 change. Know it's draft like.

server_src/constants/template.constants.js
NotExpectedYet: Yeah sounds good, will be easier once I've templated off the client side stuff.

server_src/constants/state.constants.js
NotExpectedYet: I agree with that totally.

davidzwa: Reverted

client_src/js/services/octofarm-client.service.js
NotExpectedYet: tut tut XD

server_src/constants/state.constants.js
NotExpectedYet: I do know this, just forgotten thanks for the reminder!

client_src/js/services/octofarm-client.service.js
davidzwa: Added TODO and passed errors to this overrides prop.

NotExpectedYet (approved):
LGTM, haven't tested tho

PR #728, opened 2021-08-04
Chore/various fixes for dev merge

NotExpectedYet:
@davidzwa Current Issues I have fixed for dev to get canary work. I've left these off changelog as they're a result of recent commits. 

- [x] Validation error
` UnhandledPromiseRejectionWarning: ValidationError: Printer validation failed: sortIndex: Path `sortInde
x` is required., webSocketURL: Path `webSocketURL` is required.
    at model.Document.invalidate (C:\Users\James Mackay\WebstormProjects\OctoFarm\node_modules\mongoose\lib\document
.js:2760:32)
` - This was when adding a printer. The client doesn't send this information to the server and should be generated by the printer add function. I've made them not required for now because it means re-doing the printer add function ordering. Gather you've already tackled that on your big one branch. 

- [x] Settings sends incorrect values for timeouts causing 0.03 seconds for a timeout aha. Epic. 

- [x] Websocket URL adds double `//` ie, ` wss://thinker1.corp.vexica.com//sockjs/websocket `

- [x] PrintersToConnect error when sending gcode commands in bulk...

- [x] Printers Settings doesn't remove loading icons

- [x] Updated patreon list


NotExpectedYet:
Jest has a module called `todo()` tried it on our version but didn't seem to be there. Might be worth looking into rather than skipped?

PR #732, opened 2021-08-07
fix incorrect ws url fix (#731)

NotExpectedYet:
#731 - This was a mistake on my part from switching to the big one branch.

### DIFF
--- a/server_src/runners/state.js
+++ b/server_src/runners/state.js
@@ -998,7 +998,7 @@ class Runner {
             "Complete",
             farmPrinters[i]._id
           );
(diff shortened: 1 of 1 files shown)


davidzwa:
Should we aim to use the new URL(..).href approach? Curious what your thoughts on it are, I think its kinda neat

davidzwa (changes requested):
Sorry, I think we can do better to fix either slash or not

NotExpectedYet:
> Sorry, I think we can do better to fix either slash or not

I'm not saying we can't, I wanted to get development fixed and just revert my previous changes as they we're made in error.

davidzwa (approved):
mb mb :)

PR #743, opened 2021-08-24
Bug/64 octoprint login response apikey null

davidzwa:
This PR improves apiKey checks and makes sure wrong API keys do not cause retries. It does not fix the problem itself as reproduction is required beforehand.

Scope:
- Improve OctoPrint API connection error state
- Decide not to call an endpoint when `printerState.isApiAccessible()` returns false (not on OctoPrint API level, but on task level)
- Introduce `HttpStatusCode` constant ported from V2
- Use `axios` instead of `node-fetch`
- Remove retry on OctoPrint API level - this was added on task manager level to ensure Printer State is checked for API accessibility state
- Removes app.js code to boot discovery service on startup

Future work includes:
- to remove `node-fetch` and usage from many places still using them
- to restore file downloading capabilities and add download progress with file write stream using axios
- removing retry settings OR coupling retry to the task mechanism instead

Canary stability after this PR:
- This PR introduces `httpClient` and simplifies API calls and tests.
- It can easily break code as `.json()` and `.text()` methods are not returned by `axios`.

Verdict: done concerning bug 64 resolution (DevOps), but not ready to ship as it makes canary slightly more unstable.


test/api/amialive.test.js
```diff
@@ -9,16 +9,20 @@ let request;
...
+const AxiosMock = require("../provisioning/axios.mock");
+const GithubApiServiceMock = require("../provisioning/github-api.mock");
 
 const testRoute = "/api/amialive";
 
 beforeAll(async () => {
   await dbHandler.connect();
   const { server, container } = await setupTestApp();
   testContainer = container;
+  container.register(DITokens.httpClient, awilix.asClass(AxiosMock));
```
davidzwa, line 24: `httpClient` is the `axios` instance. For now its the `require(...)` directly - this allows us to customize it later with axios instances. For now it's to first see where things are headed in the bigger picture.

server_src/tasks/printer-websocket.task.js
```diff
@@ -16,24 +22,26 @@ class PrinterWebsocketTask {
 
   #logger = new Logger("Printer-Websocket-Task");
 
-  #errorMaxThrows = 5;
-  #errorModulus = 10; // After max throws, log it every 10 failures
+  #errorMaxThrows = 3;
+  #errorModulus = 50; // After max throws, log it every x failures
```
davidzwa, line 26: Explanation: after 3 errors the logs are silenced. After each 50 retries a log is done. This ends up as ~360 logs per day at most.

server_src/tasks/printer-websocket.task.js
```diff
@@ -74,70 +82,125 @@ class PrinterWebsocketTask {
   async setupPrinterConnection(printerState) {
     const loginDetails = printerState.getLoginDetails();
     const printerName = printerState.getName();
+    const printerId = printerState.id;
```
davidzwa, line 85: Used to uniquely track error count

app.js
```diff
@@ -63,9 +63,6 @@ if (!!majorVersion && majorVersion < 14) {
       app.listen(port, "0.0.0.0", () => {
         logger.info(`Server started... open it at http://127.0.0.1:${port}`);
       });
-
-      const autoDiscoveryService = container.resolve(DITokens.autoDiscoveryService);
```
davidzwa, line 67: Stop SSDP from firing at start

test/provisioning/axios.mock.js
```diff
@@ -0,0 +1,57 @@
+class AxiosMock {
```
davidzwa, line 1: This is how easy a mock can be. It's basically a copy of the original thing without slow functionality.

- Save a response beforehand
- Call and get the saved response

davidzwa:
@NotExpectedYet This PR should give better clues what is going wrong with the api key

NotExpectedYet:
> @NotExpectedYet This PR should give better clues what is going wrong with the api key

The key comes back null from OP itself on mine. Think that's why I hit the settings endpoint in the first place but will take a look when I can and try it out.

davidzwa:
> > @NotExpectedYet This PR should give better clues what is going wrong with the api key
> 
> The key comes back null from OP itself on mine. Think that's why I ended up hitting the settings endpoint in the first place but will take a look when I can and try it out as gather it never does on your instances.

If the response contains `apikey: null` that indicates a global API key. Tested with User API key only, so will test with Application Key as well.

NotExpectedYet:
> > > @NotExpectedYet This PR should give better clues what is going wrong with the api key
> > 
> > 
> > The key comes back null from OP itself on mine. Think that's why I ended up hitting the settings endpoint in the first place but will take a look when I can and try it out as gather it never does on your instances.
> 
> If the response contains `apikey: null` that hints at a global API key. Tested with User API key only, so will test with Application Key as well.

Yeah mine are application keys.

davidzwa:
> > > > @NotExpectedYet This PR should give better clues what is going wrong with the api key
> > > 
> > > 
> > > The key comes back null from OP itself on mine. Think that's why I ended up hitting the settings endpoint in the first place but will take a look when I can and try it out as gather it never does on your instances.
> > 
> > 
> > If the response contains `apikey: null` that hints at a global API key. Tested with User API key only, so will test with Application Key as well.
> 
> Yeah mine are application keys.

I just double checked. Both User and App API keys pass correctly for me locally. Global API Key is marked as Global correctly as well.

NotExpectedYet (approved):

Issue #647, opened 2021-06-24
Docker support for arm64 (architecture request)


cranky-coder:
### OctoFarm Version

latest

### OctoPrint Version(s)

1.6

### pm2 or docker?

docker

### Operating System

Raspberry pi os 64, ubuntu arm 64

### Location of the problem

Not sure

### Description of the problem

```bash
Curious if the dockerbuild could be updated to include arm64 architecture.  Or if there is something that is preventing arm64 support.  I have a bunch of pi4's here running a kubernetes cluster and interested in seeing if i could get octofarm running in it.
```

### Steps taken to reproduce

```bash
Docker pull on pi os 64 doesn't run.
```

### What did you expect to happen

```bash
NA
```

### To avoid the problem I had to

```bash
NA
```

davidzwa:
Hi @cranky-coder, we're currently working on an image called FarmPi.
Visit the discord for the way to get a beta test download:
https://discord.gg/G6cCkqVV

For the rest we might start testing/emulating ARM in future as we saw that indeed the Raspberry Pi is a bit of a underdog right now for OctoFarm. This requires time which is precious. Does the raspberry pi image above act as an acceptable substitute for you for the meanwhile?

NotExpectedYet:
Did you try the alpine image?

davidzwa:
@cranky-coder we're awaiting your response.

cranky-coder:
On the dockerhub https://hub.docker.com/r/octofarm/octofarm/tags

the alpine image is listed as linux/amd64.  Not arm64.  So as far as I can tell, switching to that alpine image will still not give me the arm64 architecture docker support.

In the previous response, you ask "if the above raspberry pi image above act as acceptable substitute"  but I see no mention of pi image.  But I am looking for a docker image support on the arm64 architecture to deploy the octofarm control system into a kubernetes environment built on pi4 64bit modules.  That is why I was inquiring on the arm64 docker specifically.   A custom pi image at this time would not change anything.  I was just curious if the ci/cd process for the docker image could be adjusted to also add the arm64 instead of single arch.

Maybe with something like the docker buildx multi-arch
https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

davidzwa:
A user posted this
https://github.com/OctoFarm/OctoFarm/discussions/573

cranky-coder:
That is for arm32.  Won't work a 64 bit raspberrry pi setup.  

[...]

PR #795, opened 2022-01-28
Build images for arm v7 and arm64

jwkicklighter:
This is untested because I'm not sure the best way to run the action on my own fork, but after reading #647 and investigating `package.json`, `Dockerfile`, the relevant entrypoint file, and the action workflow, I believe that specifying the platforms here may be all that is left to have ARM builds.


NotExpectedYet:
Hi, could this be changed over to the development branch please? 

The nodejs 15 check will pass then.

Thanks for looking into this though! The only one I'm not so sure about is the monolithic but they'll be able to build on the development tag to be tested.

NotExpectedYet:
> This is untested because I'm not sure the best way to run the action on my own fork, but after reading #647 and investigating `package.json`, `Dockerfile`, the relevant entrypoint file, and the action workflow, I believe that specifying the platforms here may be all that is left to have ARM builds.


Success! Realized it didn't matter about mongodb. 

Big thanks again!

image

jwkicklighter:
@NotExpectedYet thanks for getting that onto the right branch! I actually looked at the branch structure before I forked, but apparently I wasn't awake enough when I made the PR. Glad to see the build go through, hopefully there are no surprises!

NotExpectedYet:
@jwkicklighter 

Gives us some breathing room been on development at least. I'll make sure to give them a check out but should be no issues. Thanks again!

